PR #6, opened 2018-02-08
Adding Ignore Errors Support

tance77:
I believe this should add the support to ignore certain errors as defined here: https://docs.sentry.io/clients/javascript/config/

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -10,6 +10,7 @@ export interface IElectronSentryOptions {
   companyName?: string;
   environment?: string;
   tags?: any;
+  ignoreErrors?: any;
 }
 
 export const defaults: IElectronSentryOptions = {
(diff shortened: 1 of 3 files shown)


src/options.ts
```diff
@@ -10,6 +10,7 @@ export interface IElectronSentryOptions {
   companyName?: string;
   environment?: string;
   tags?: any;
+  ignoreErrors?: any;
```
timfish: I think to allow only string and regex input the types need to be:

```
ignoreErrors?: (string | RegEx)[];
```

timfish:
I'm away from my desk for a while so will not be able to properly check this for a while!

tance77:
I went ahead and made those changes. I have no idea how to use typescript so I was just winging it and using your code as an example.

timfish:
This will be possible with the @sentry/core API changes. Docs and release to come.

Issue #1, opened 2018-02-06
Extra Context


tance77:
Is it possible to set  [Extra context.](https://docs.sentry.io/learn/context/) If it is possible can you show an example?

timfish:
I don't think this is currently possible without calling into `raven-js` and `raven-node` yourself.

This will be fixed by #2

Issue #2, opened 2018-02-07
Follow Sentry SKD guidelines


timfish:
Once Sentry publish the repositories for [@sentry/core](https://www.npmjs.com/package/@sentry/core) and [@sentry/browser](https://www.npmjs.com/package/@sentry/browser) I plan to refactor to use these and copy their API's. 

This would bring this package in line with the Sentry SDK guidelines which aren't currently being followed...

PR #7, opened 2018-02-23
ref: Migrate to Sentry SDK Core

jan-auer:
This PR uses pre-release versions the new `@sentry/core`, `@sentry/node` and `@sentry/browser` SDKs in place of the Raven* client libraries. Besides that, this PR brings the code base in line with the other SDK previews and updates the linter configuration.

One major change is the way this sends events now. It no longer uses the Electron `CrashReporter` to upload minidumps but submits them manually. This allows us to persist Breadcrumbs and Context information beyond application crashes and attach them directly to native crash reports.

I've checked in a small example to test all sorts of crash events. A test suite, more advanced features and a more stable uploader are following in future PRs.


jan-auer:
Forgot to update the readme, but we can do that afterwards. There is some description in the comment on the `SentryElectron` class. 

Also please note that the SDK namespaces are still not final. I think the naming of some of the classes is a bit odd in some cases, so we might consider to change that before pushing out a new release.

package.json
```diff
@@ -1,28 +1,40 @@
 {
-  "name": "electron-sentry",
-  "description": "JavaScript and native crash reporting from Electron to Sentry",
-  "version": "1.1.2",
```
timfish: This will need to be greater than the last to publish to npm. 

As the entire API has changed v2 would be preferable!

package.json
```diff
@@ -1,28 +1,40 @@
...
+  "name": "@sentry/electron",
+  "description": "Offical Sentry SDK for Electron",
+  "version": "0.1.0",
```
timfish: This will need to be greater than the last to publish to npm.

As the entire API has changed v2 would be preferable!

timfish:
Closes #1, #2

package.json
timfish: Or do you guys want to just publish to `@sentry/electron`?

jan-auer: That would be the idea. We could experiment a little more with the API before making it stable then

src/lib/electron.ts
```diff
@@ -0,0 +1,501 @@
...
+  }
+
+  /** Returns whether JS is enabled and we are in a renderer process. */
+  private isRenderEnabled(): boolean {
+    return this.isRenderProcess() && this.options.enableJavaScript !== false;
+  }
+
+  /** Returns whether native reports are enabled. */
+  private isNativeEnabled(): boolean {
+    // On macOS, we should only start the Electron CrashReporter in the main
```
timfish: The electron docs seem to contradict this:

>Note: On macOS, Electron uses a new crashpad client for crash collection and reporting. If you want to enable crash reporting, initializing crashpad from the main process using crashReporter.start is required regardless of which process you want to collect crashes from. Once initialized this way, the crashpad handler collects crashes from all processes.

timfish: Or is that simply so the crashes are sent to the server?

jan-auer: Crashpad spawns a new process and attaches to the Electron processes to watch them (required since macOS Mavericks). This is why it should only be necessary to start the crash reporter in the main process. It seems that this is also what the Electron docs try to express. 
On Windows and Linux, Electron still uses Breakpad. In that case, it should be necessary to start the CrashReporter in every process. 
So far, I have only tested thoroughly on macOS.

timfish:
Thanks @jan-auer and @HazAT, this is looking great!

For tests I intended to add [Spectron](https://electronjs.org/spectron) e2e tests and have them run in CI for all three platforms. They would spawn up local server to submit to, [much like the Electron tests](https://github.com/electron/electron/blob/master/spec/api-crash-reporter-spec.js) do. I've got Spectron tests working in CI for a couple of Electron apps.

HazAT:
@timfish Yeah that was also our idea, we already discussed it but wanted to merge this first, test setup was the next step.
As you said we thought about adding mocha with spectron, ideally the tests should have the same "style" as the one in our core/browser package.

https://github.com/getsentry/raven-js/blob/4.0.0-wip/packages/browser/test/index.test.ts

src/lib/store.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Creates a new store.
+   *
+   * @param filename A unique filename to store this data.
+   * @param initial An initial value to initialize data with.
+   */
+  constructor(filename: string, private initial?: T) {
+    if (!existsSync(BASE_PATH)) {
+      mkdirSync(BASE_PATH);
```
timfish: This causes an exception if its the first time the app has been run as the base `AppData` directory does not exist.

HazAT:
@timfish Also do you mind transferring the repo to our org (@getsentry) we'll add you as an admin but then our team could start to contribute easier.
We also want to rename it to `sentry-electron` to be consistent with other SDK repos.

src/lib/store.ts
timfish: `AppData` is not created until `makeSingleInstance` is called or you create `BrowserWindows` and a few other assorted APIs.

HazAT: Would this be solved when we lazy create the folder?

timfish:
Tried to transfer ownership but got:
```
You don’t have the permission to create repositories on getsentry
```

HazAT:
@timfish OK, can you add me as an admin, I will do it.

src/lib/store.ts
timfish: Somebody could write an Electron app where they don't hit any points that create the `AppData` directory. One example I can think of is an app which starts in the tray and doesn't create any windows immediately. It's safe to check if `AppData` exists and create it if not.

It'll be worth noting in the docs that if you change the location of `AppData`, this needs to be done before starting Sentry. On Windows it common to change it to `%LOCALAPPDATA%\AppName` so you don't fill up roaming profiles.

src/lib/electron.ts
```diff
@@ -0,0 +1,501 @@
...
+    );
+  }
+
+  /** Activates the Electron CrashReporter. */
+  private async installNativeHandler(): Promise<boolean> {
+    // We will manually submit errors, but CrashReporter requires a submitURL in
+    // some versions. Also, provide a productName and companyName, which we will
+    // add manually to the event's context during submission.
+    crashReporter.start({
+      productName: app.getName(),
```
timfish: This fails in the renderer, it should be `(app || remote.app)`

HazAT:
@timfish I've changed it so we create the AppData folder if it does not exist, should we merge this now and create the next PR for the test setup?

timfish:
Yeah, go for it.

PR #11, opened 2018-02-24
feat: Normalise paths

### LINKED ISSUE #9: URLs should be normalised to remove full path
Like [previously](https://github.com/getsentry/sentry-electron/blob/31312ce0ef91aaed2c8e03e154e9aee8bbdd1c05/src/electron-sentry.ts#L276), URLs need normalising to the app base so Sentry can group them correctly.

The previous regex only catered for apps packaged in asar.

timfish:
Closes #9

### DIFF
--- a/src/lib/electron.ts
+++ b/src/lib/electron.ts
@@ -513,8 +513,7 @@ export class SentryElectron implements Adapter {
     }
 
     if (event.request && event.request.url) {
-      event.request.url =
-        this.normalizeUrl(event.request.url);
+      event.request.url = this.normalizeUrl(event.request.url);
     }
 
     const stacktrace =
--- a/src/lib/electron.ts
+++ b/src/lib/electron.ts
@@ -211,6 +211,11 @@ export class SentryElectron implements Adapter {
    */
   public async captureBreadcrumb(breadcrumb: Breadcrumb): Promise<Breadcrumb> {
     if (this.isRenderProcess()) {
+      if (breadcrumb.category && breadcrumb.category === 'navigation') {
+        breadcrumb.data.from = this.normalizeUrl(breadcrumb.data.from);
+        breadcrumb.data.to = this.normalizeUrl(breadcrumb.data.to);
+      }
+
       ipcRenderer.send(IPC_CRUMB, breadcrumb);
       return breadcrumb;
     }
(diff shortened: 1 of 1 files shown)


src/lib/electron.ts
```diff
@@ -498,4 +500,44 @@ export class SentryElectron implements Adapter {
...
+      event.culprit = this.normalizeUrl(event.culprit, appPath);
+    }
+
+    if (event.request && event.request.url) {
+      event.request.url =
+        this.normalizeUrl(event.request.url.replace('file:///', ''), appPath.replace(/\\/g, '/'));
+    }
+
+    const stacktrace =
+      event.stacktrace
```
timfish: Is `event.stacktrace` to find the stacktrace in older browsers? Is this even required in Electron?

src/lib/electron.ts
```diff
@@ -498,4 +500,44 @@ export class SentryElectron implements Adapter {
...
   }
+
+  private normalizeEvent(event: any) {
+    const appPath = app.getAppPath();
+
+    if (event.culprit) {
+      event.culprit = this.normalizeUrl(event.culprit, appPath);
+    }
+
+    if (event.request && event.request.url) {
```
timfish: Is `event.request.url` normalisation required to properly fingerprint in Sentry?

HazAT: If there is a stacktrace, the the frames of it will be used for the fingerprint, so I would keep it like this.

HazAT:
BTW
Feel free to clone this repo and create PRs directly here. In case we want to add commits to your branch.

HazAT:
https://forum.sentry.io/t/replacing-culprits-with-transaction-names/105
`culprit` is called transaction now

timfish:
@HazAT This option is ticked by default so you should still be able to add commits to my branch: 

image

timfish:
It looks like both `@sentry/node` and `@sentry/browser` are calling `send()` with events containing `culprit` so it needs fixing in both of those and then here?

timfish:
This is going to require more work. This current normalising works on the example app but fails on a couple of production apps I've tested. 

I think this difference is loading JavaScript source through `require` (ie. nodejs) and through HTTP requests. 

When I run the example app and trigger a JavaScript exception in the renderer process I get:
```
culprit: 'C:\\Users\\tim\\Documents\\Repos\\sentry-electron\\example\\renderer.js'
```
When I do the same in a customers app in dev mode I get:
```
culprit: 'file:///C:/Users/tim/Documents/Repos/PicoLog6/dist/layout/dev-menu.js',
```

timfish:
@HazAT are you happy with this as it is for now?

We just need to remember to include tests that throw exceptions in resources loaded via both `require` and `file://` as the paths vary.

PR #12, opened 2018-02-24
feat: Add option to clear breadcrumbs from previous sessions

### LINKED ISSUE #10: Persisted breadcrumbs should be cleared when app starts
Breadcrumbs are persisted to disk so that they are available on next run in case of native crash in main process. 

Currently, all events posted to Sentry contain breadcrumbs from multiple previous app sessions.

timfish:
Fixes #10 

Should the context get cleared too?

### DIFF
--- a/src/lib/electron.ts
+++ b/src/lib/electron.ts
@@ -375,7 +375,7 @@ export class SentryElectron implements Adapter {
       release: this.options.release,
       environment: this.options.environment,
       sdk: { name: SDK_NAME, version: SDK_VERSION },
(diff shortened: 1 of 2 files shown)


timfish:
Yep, that makes sense. 

I will make the clearing optional but shall I leave `Array.from(this.breadcrumbs.get())` there? At startup `sendNativeCrashes` is not awaited on so if developers decide to clear breadcrumbs at startup they might clear them before the main process native crash is sent.

One thing I noticed was that its not entirely clear when each session starts and ends so we might have to add `app.on('quit', ...` and flush the stores.

PR #19, opened 2018-02-26
feat: Add tests with spectron

HazAT:


.travis.yml
```diff
@@ -20,3 +26,11 @@ matrix:
     - env: NAME="lint"
       script:
         - yarn lint
+    - env: NAME="test"
+      install:
+        - export DISPLAY=':99.0'
+        - sh -e /etc/init.d/xvfb start
+        - sleep 3
```
timfish: There's the `xvfb-maybe` module which can make this simpler.

You can exclude all the above and then:
```
  "test": "xvfb-maybe mocha ./dist/test"
```
And then it will sort out xvfb if its ever required

HazAT: thx :)

HazAT:
I merged this, even though the e2e tests actually don't really test something useful yet.

PR #20, opened 2018-02-26
fix: Remove electron peerDependency

### LINKED ISSUE #17: Electron should not be in peerDependencies
When creating an electron project, electron it is added to `devDependencies`. This is because the module it's only used when running in development or packaging a production app using `electron-builder` or `electron-packager`.

Including electron in `peerDependencies` means the `electron-builder` or `electron-packager` package electron as a dependency in `node_modules` which means the electron binaries get included twice. 😃

timfish:
Closes #17

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,9 +23,6 @@
     "node-fetch": "^2.0.0",
     "util.promisify": "^1.0.0"
   },
-  "peerDependencies": {
-    "electron": "*"
-  },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


timfish:
```
ERROR: src/lib/electron.ts[22, 8]: Module 'electron' is not listed as dependency in package.json
ERROR: src/lib/store.ts[4, 29]: Module 'electron' is not listed as dependency in package.json
```

Issue #16, opened 2018-02-26
Offline/Retry 


HazAT:
We should locally cache events if the user has no internet connection.
Also if the server replies with 429.
See https://github.com/getsentry/sentry-cocoa for details

PR #25, opened 2018-02-28
feat: Offline/Retry support for minidumps

kamilogorek:
Ref: https://github.com/getsentry/sentry-electron/issues/16

Heavy work in progress


src/lib/uploader.ts
```diff
@@ -94,6 +99,13 @@ export default class MinidumpUploader {
         // We either succeeded or something went horribly wrong
         // Either way, we can remove the minidump file
         await unlink(request.path);
+
+        // Also remove it from queued minidumps
+        const storedRequests = this.queue.get();
+        storedRequests.splice(storedRequests.indexOf(request), 1);
```
kamilogorek: This will blow up if the event is somehow not found in the queue. It'll do `splice(-1, 1)`, which removes last element.

I'd go with
```js
this.queue.set((storedRequests) => storedRequests.filter(storedRequest => storedRequest !== request));
```

src/lib/uploader.ts
```diff
@@ -135,6 +147,18 @@ export default class MinidumpUploader {
     });
   }
 
+  /**
+   * Flushes locally cached minidumps from the queue.
+   */
+  public async flushQueuedMinidumps(): Promise<void> {
+    const storedRequests = this.queue.get();
+    if (storedRequests.length) {
```
kamilogorek: Queue will always be an array, so there's no need to check if there's anything. It'll noop for empty list.

HazAT:
@kamilogorek @timfish Can we merge this so we can add tests in a later PR?
There is more in this PR than only minidump retry (code reformating/linting)

kamilogorek:
Done.

timfish:
The Travis tests never ran and I checked their service status and it said
something about emergency maintenance.

PR #34, opened 2018-03-12
ref: Update to latest sentry/core version

jan-auer:
- [x] Update TS and TSLint configurations
- [x] Fix errors in utils and `MinidumpUploader`
- [x] Implement `ElectronBackend`
- [x] Implement `ElectronFrontend` and `SentryClient`
- [x] Update and fix tests
- [x] Move generic utils to `@sentry/utils` (`filterAsync` and `clone`)
- [x] Update the readme


src/lib/backend.ts
```diff
@@ -0,0 +1,457 @@
...
+    await this.frontend.setOptions({
+      onFatalError: async (error: Error) => {
+        console.error('*********************************');
+        console.error('* SentryElectron unhandledError *');
+        console.error('*********************************');
+        console.error(error);
+        console.error('---------------------------------');
+        await this.frontend.captureException(error);
+      },
+      ...this.frontend.getOptions(),
```
jan-auer: I think we can remove this thanks to https://github.com/getsentry/raven-js/blob/4.0.0-wip/packages/core/src/lib/base.ts#L198

PR #35, opened 2018-03-13
fix: Stop using Spectron to run e2e tests

timfish:
- [x] Removed e2e stuff from example app
- [x] Added test app and fixtures
- [x] Simplified the test context
- [x] Fix Linux tests


timfish:
```
    ✓ Native crash in renderer process (10703ms)
```
It takes about 5s on my Mac still too. My guess is that the renderer `crashed` event is slow on Mac?

PR #43, opened 2018-03-16
feat: Test paths with spaces

timfish:
- Move fixtures inside app basepath so paths get normalised
- Add checks for normalised paths
- Add a failing test for path with spaces


test/e2e/test-app/fixtures/javascript main with spaces.js
```diff
@@ -0,0 +1,3 @@
+if (process.type === 'browser') {
+  throw new Error('Error triggered in renderer process');
```
jan-auer: `process.type === 'browser'` indicates the main process.

timfish: Yeah crazy isn't it. I think the decision was made early on and now they can't change it.

timfish: But yeah, the error text are wrong...

PR #46, opened 2018-03-16
Fix: onFatalError

### LINKED ISSUE #45: onFatalError option doesn't override the default
```javascript
    SentryClient.create({
      dsn: '__DSN__',
      onFatalError: (error) => {
        console.log('Something!');
      }
    })
```
Only the default [gets set](https://github.com/getsentry/sentry-electron/blob/d49e0b11b8d0e67f5c921f8de022d1576d5415ff/src/lib/backend.ts#L343)
```
*********************************
* SentryElectron unhandledError *
*********************************
Error: Testing
..snip
```

timfish:
Fixes #45 

Also added a renderer `unhandledrejection` test

### DIFF
--- a/src/lib/backend.ts
+++ b/src/lib/backend.ts
@@ -339,14 +339,23 @@ export class ElectronBackend implements Backend {
 
   /** Activates the Node SDK for the main process. */
   private async installMainHandler(): Promise<boolean> {
+    const options = this.frontend.getOptions();
+    // tslint:disable-next-line:no-unbound-method
+    const customHandler = options.onFatalError;
+
     await this.frontend.setOptions({
       onFatalError: async (error: Error) => {
-        console.error('*********************************');
-        console.error('* SentryElectron unhandledError *');
-        console.error('*********************************');
-        console.error(error);
-        console.error('---------------------------------');
         await this.frontend.captureException(error);
+
+        if (customHandler) {
+          customHandler(error);
+        } else {
+          console.error('*********************************');
+          console.error('* SentryElectron unhandledError *');
+          console.error('*********************************');
+          console.error(error);
+          console.error('---------------------------------');
+        }
       },
     });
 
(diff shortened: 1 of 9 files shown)


src/lib/backend.ts
```diff
@@ -339,14 +339,23 @@ export class ElectronBackend implements Backend {
 
   /** Activates the Node SDK for the main process. */
   private async installMainHandler(): Promise<boolean> {
+    const options = this.frontend.getOptions();
+    // tslint:disable-next-line:no-unbound-method
+    const customHandler = options.onFatalError;
```
timfish: @jan-auer is there a way to do this without disabling `no-unbound-method`?

timfish:
Need to investigate the flakey test

src/lib/backend.ts
jan-auer: You could trick the linter for now by writing:
```ts
    const { onFatalError } = this.frontend.getOptions();
```

src/lib/backend.ts
```diff
@@ -339,14 +339,23 @@ export class ElectronBackend implements Backend {
...
+
     await this.frontend.setOptions({
       onFatalError: async (error: Error) => {
-        console.error('*********************************');
-        console.error('* SentryElectron unhandledError *');
-        console.error('*********************************');
-        console.error(error);
-        console.error('---------------------------------');
         await this.frontend.captureException(error);
```
jan-auer: It is a good idea to always log the error. However, this implementation makes it hard to understand what is happening due to its side effects. I'd rather prefer a simpler (though not equal) fix for now, as it was originally intended:

```ts
    if (!this.frontend.getOptions().onFatalError) {
      await this.frontend.setOptions({
        onFatalError: async (error: Error) => {
          console.error('*********************************');
          console.error('* SentryElectron unhandledError *');
          console.error('*********************************');
          console.error(error);
          console.error('---------------------------------');
          await this.frontend.captureException(error);
        },
      });
    }
```

PR #50, opened 2018-03-19
fix: Breadcrumbs

### LINKED ISSUE #41: getPath('userData') should only be evaluated when calling create
[`const CACHE_PATH = join(getApp().getPath('userData'), 'sentry');`](https://github.com/getsentry/sentry-electron/blob/8a0f101af8bcd2a47d75b5edc035394d89811fc2/src/lib/backend.ts#L41)

Gets evaluated when the library is required/imported rather than when create is called. 

If developers change the `userData` directory after importing but before calling `create` the cache path can be incorrect.

timfish:
- Only store breadcrumbs/context in main process
- Send breadcrumbs/context over IPC from renderer
- Also Fixes #41


src/lib/frontend.ts
```diff
@@ -43,6 +54,28 @@ export class ElectronFrontend extends FrontendBase<
     const prepared = await this.prepareEvent(event);
     await this.getBackend().uploadMinidump(path, prepared);
   }
+
+  /**
+   * @inheritDoc
+   */
+  public async addBreadcrumb(breadcrumb: Breadcrumb): Promise<void> {
+    if (isRenderProcess()) {
```
jan-auer: We should do the same for `ElectronFrontend.captureEvent` and instead add an assert in `ElectronBackend.sendEvent`. I just realized that otherwise an event goes through the lifecycle hooks twice (once in the renderer and once in the main process).

src/lib/backend.ts
```diff
@@ -121,9 +104,10 @@ export class ElectronBackend implements Backend {
    */
   public async install(): Promise<boolean> {
     let success = true;
+    const cachePath = join(getApp().getPath('userData'), 'sentry');
```
jan-auer: This could go in a utility `getCachePath()` so that we can call it more easily from different places.

src/lib/backend.ts
```diff
@@ -160,14 +147,16 @@ export class ElectronBackend implements Backend {
    * @inheritDoc
    */
   public async storeContext(context: Context): Promise<void> {
-    this.context.set(context);
+    if (this.context) {
+      this.context.set(context);
```
jan-auer: Could you throw a `SentryError` here and in `loadContext` instead? We really don't want to call these in the renderer and don't want to miss it in any case. See `sendNativeCrashes` for an example.

timfish: Yeah, sure. Wasn't sure whether they got called by `core` in the renderer.

src/lib/frontend.ts
timfish: 👍

PR #60, opened 2018-04-10
fix: use private DSN in docs

### LINKED ISSUE #58: Should you use private or public DSN?
Just saw the updated documentation on sentry website for electron integration.

Initally I thought that you must use private DSN. But in the new doc, the example uses public DSN.
Is it ok to use public DSN for main an renderer process? Is there any downside?

timfish:
Fixes #58


jan-auer:
See reasoning in https://github.com/getsentry/sentry-electron/issues/58#issuecomment-380092975

PR #69, opened 2018-04-18
feat/test-multiple-versions

timfish:


timfish:
Need to work out why it fails on Mac

timfish:
There is still come context leaking between tests. This is because every test uses the same crashes directory `test-app Crashes`.

PR #78, opened 2018-04-30
feat: Use electron's net module as http transport

### LINKED ISSUE #73: HTTPS requests are blocked on corporate networks
Many corporate networks unwrap SSL connections and install their own self signed root certificate into the Windows certificate store. 

Requests made from the main process do no respect the users proxy settings or check certificates against those outside the certificates shipped with node.

It looks like the [`win-ca`](https://www.npmjs.com/package/win-ca) module could fix this this on Windows but requires native build tools to build it for the current platform.

jan-auer:
Requires https://github.com/getsentry/raven-js/pull/1316
Fixes #73


jan-auer:
Note this won't build until we merge the linked PR.

package.json
```diff
@@ -35,8 +35,9 @@
     "@sentry/node": "0.5.2",
     "@sentry/shim": "0.5.2",
     "@sentry/utils": "0.5.2",
+    "@types/node-fetch": "^1.6.9",
```
timfish: Is it not enough for this to be in `devDependencies`?

jan-auer:
Gonna merge this now, Windows test verified locally.

PR #79, opened 2018-05-03
ref: Separate implementation for processes

jan-auer:
This PR separates the SDK implementation for renderers from the main process. Virtually no code was shared anymore, so it makes more sense to keep everything separate. 

In order to maintain a common interface to the outside user, there are common interfaces that need to be implemented on both sides:

 - `CommonFrontend` extends `Frontend` and defines an additional captureMinidump method
 - `CommonBackend` is a re-export for `Backend` for now, and might receive additional backend methods at some point. 

Furthermore, there is a a common `ElectronFrontend` class that lazily requires either `MainFrontend` or `MainBackend` and dispatches all method calls there. This proxy class is needed for type safety of the outside user.

With these changes, all dependencies of the `renderer` module comply with the Electron sandbox mode now and should also work with disabled `node_integration`.


src/renderer/frontend.ts
```diff
@@ -0,0 +1,81 @@
...
+   * @param path The relative or absolute path to the minidump.
+   * @param event Optional event payload to attach to the minidump.
+   * @param scope The SDK scope used to upload.
+   */
+  public async captureMinidump(
+    _path: string,
+    _event: SentryEvent,
+    _scope: Scope,
+  ): Promise<void> {
+    // TODO: Figure out whether this makes sense
```
jan-auer: Implementing this is required by the common interface, but it doesn't really make sense to have this in the renderer. The Electron crashReporter does not allow to create dumps on-the-fly and renderers _might_ not have access to the file system anyway. So the default way to handle this would be an IPC to the main process. 

Now, we can issue an IPC call here, but I'm questioning how useful that would really be. The other alternative would be to completely remove the `CommonFrontend` interface and instead have fully separate SDKs, e.g. `@sentry/electron/main` and `@sentry/electron/renderer`. That comes with a whole range of different problems, though.

jan-auer:
@timfish Thanks for that point, will do some more extensive testing in a follow-up PR 👍

Issue #81, opened 2018-05-12
I update to 0.5.4, error on MODULE_NOT_FOUND


ncaq:
I update `@sentry/electron` 0.4.1 to 0.5.4.

And I rewrite function create to init.

My `src/sentry.js`.

~~~
import { init, captureMessage } from '@sentry/electron';

init({ dsn: 'foobar' });
captureMessage('Hello, world!');
~~~

So sentry is failed.
(My application name replaced to foo)

~~~
{ Error: Cannot find module "/home/ncaq/Desktop/foo/package.json".
    at webpackEmptyContext (/home/ncaq/Desktop/foo/app/webpack:/node_modules/@sentry/electron/dist/main sync:2:1)
    at getPackageJson (/home/ncaq/Desktop/foo/app/webpack:/node_modules/@sentry/electron/dist/main/context.js:80:1)
    at /home/ncaq/Desktop/foo/app/webpack:/node_modules/@sentry/electron/dist/main/context.js:240:25
    at Generator.next (<anonymous>)
    at fulfilled (/home/ncaq/Desktop/foo/app/webpack:/node_modules/@sentry/electron/dist/main/context.js:4:42)
    at <anonymous> code: 'MODULE_NOT_FOUND' }
~~~

However, `/home/ncaq/Desktop/foo/package.json` exist.
Since this is a full path, it is hard to think that searching for webpack or babel is the cause.

Is it bug?
Or my setting is invalid?

---

My env.

* babel: 7.0.0-beta.46
* `@sentry/electron`: 0.5.4
* react: 16.2.0
* webpack: 4.6.0
* electron: 1.8.4
* OS: Gentoo Linux
* node: v9.10.1

jan-auer:
The reason for this error is that webpack replaces `require`, which is required by the SDK to load your app's package.json. Can you please try out the fix from #82 and see whether it works now?

```bash
npm install 'https://github.com/getsentry/sentry-electron#fix/webpack-require'
```

maplemike:
I followed the instructions and get the following error now instead of module not found:

>TypeError: e.require is not a function
    at a (/xxx/app/webpack:/sentry-electron/dist/main/context.js:83:25)
    at /xxx/app/webpack:/sentry-electron/dist/main/context.js:2:45
    at Generator.next (<anonymous>)
    at i (/xxx/app/webpack:/sentry-electron/dist/main/context.js:6:33)
    at <anonymous>

jan-auer:
[...]

PR #84, opened 2018-05-14
fix: Dispatch correctly for bundlers like webpack

jan-auer:
Frontend bundlers like webpack and rollup will statically resolve all `require`s, which might break the build or switch to the main process implementation in renderers. This PR should fix this by using `module.require` for importing the main implementation, which bypasses static bundlers entirely.

The updated example also verifies that sandbox mode works. Next step would be to update tests to use `sandbox: true` and then remove the disclaimer from docs and add a guide on how to use the SDK in sandbox mode properly, instead.

This diff is rather large, see the comment below for the relevant section. There should be sufficient code documentation above that statement explaining the reasoning behind this approach.

See #81 for more information.


src/dispatch.ts
```diff
@@ -14,14 +14,30 @@ export class ElectronFrontend implements CommonFrontend {
...
+    // implementation, which should be fine for most cases. False positives of
+    // this would be running `@sentry/electron` in a bare node process, which is
+    // acceptable.
+
     // tslint:disable:no-var-requires no-unsafe-any
-    const frontendClass =
+    const frontendClass: FrontendClass<CommonFrontend, ElectronOptions> =
       process.type === 'browser'
-        ? require('./main').MainFrontend
+        ? module.require('./main').MainFrontend
```
jan-auer: This is the actual fix, the rest is just an updated example.

jan-auer:
This is ready now.

PR #89, opened 2018-07-02
feat: Use 4.0.0

HazAT:
This PR uses the `4.0.0-beta.x` packages internally now which mostly causes internal refactorings/rewrites.
There is a breaking change with context/scope handling which is documented in the changelog / docs.

On `uncaughtException` if not in `prod` we show the dialog the same way as electron does, if in `prod` we exit the app if the user does not overwrite this behavior with `onFatalError` option.


src/main/backend.ts
```diff
@@ -53,38 +53,41 @@ function getCachePath(): string {
...
     crashed_url: normalizeUrl(contents.getURL()),
   };
 }
 
+/**
+ * Retruns a promise that resolves when app is ready.
+ */
+export async function isAppReady(): Promise<boolean> {
+  return app.isReady()
+    ? Promise.resolve(true)
```
HazAT: nit: Write `true`

HazAT:
@timfish Could you please do the last pass over it whenever you have time?
I already reviewed it with @jan-auer and it should be fine.

src/main/backend.ts
```diff
@@ -65,7 +65,7 @@ function getRendererExtra(
  */
 export async function isAppReady(): Promise<boolean> {
   return app.isReady()
-    ? Promise.resolve(true)
+    ? true
```
jan-auer: _nit:_ `app.isReady() || new Promise(...);`

src/main/backend.ts
```diff
@@ -391,19 +379,16 @@ export class MainBackend implements CommonBackend {
...
-    const event: SentryEvent = {
-      breadcrumbs: storedScope.getBreadcrumbs(),
-      extra: nextExtra,
-      tags: storedScope.getTags(),
-      user: storedScope.getUser(),
-    };
-    // tslint:enable:no-unsafe-any
+    const currentCloned = Scope.clone(getDefaultHub().getScope());
+    const fetchedScope = this.scopeStore.get();
+    (fetchedScope as any).eventProcessors = [];
```
jan-auer: This looks a bit fishy. The stored scope should never have eventProcessors. It might be more defensive to remove the eventProcessors when setting into the store instead of when fetching it...

src/main/normalize.ts
```diff
@@ -56,19 +57,18 @@ function getStacktrace(event: SentryEvent): Stacktrace | undefined {
...
   // NOTE: Events from Raven currently contain data that does not conform with
   // the `SentryEvent` interface. Until this has been resolved, we need to cast
   // to avoid typescript warnings.
-  // const copy = clone(event);
-
+  const copy = clone(event);
   // The culprit has been deprecated about two years ago and can safely be
   // removed. Remove this line, once this has been resolved in Raven.
-  delete (event as { culprit: string }).culprit;
+  delete (copy as { culprit: string }).culprit;
```
jan-auer: This should not be necessary anymore with the new Node SDK?

src/main/backend.ts
```diff
@@ -53,38 +53,41 @@ function getCachePath(): string {
...
+  private readonly scopeStore: Store<Scope>;
 
   /** Uploader for minidump files. */
   private uploader?: MinidumpUploader;
 
   /** Creates a new Electron backend instance. */
-  public constructor(frontend: Frontend<ElectronOptions>) {
-    this.frontend = frontend;
-    this.inner = new NodeBackend(frontend);
+  public constructor(private readonly options: ElectronOptions) {
```
jan-auer: _nit_: This is the only time we declare fields like this.

src/main/client.ts
```diff
@@ -0,0 +1,76 @@
...
+   * @param scope Optional SDK scope used to upload.
+   */
+  public async captureMinidump(
+    path: string,
+    event: SentryEvent = {},
+    scope?: Scope,
+  ): Promise<void> {
+    event.tags = { event_type: 'native', ...event.tags };
+    await this.processEvent(
+      event,
```
jan-auer: _nit_: In future refactoring, consider: `{ ...event, tags: { event_type: 'native', ...event.tags } }`

timfish:
I'll take a look and test this with our app on Monday

src/main/integrations/onuncaughtexception.ts
```diff
@@ -0,0 +1,57 @@
...
+      getDefaultHub().withScope(async () => {
+        getDefaultHub().addEventProcessor(async (event: SentryEvent) => ({
+          ...event,
+          level: Severity.Fatal,
+        }));
+
+        const nodeClient = getDefaultHub().getClient() as NodeClient;
+        await nodeClient.captureException(error, getDefaultHub().getScope());
+
+        if (process.env.NODE_ENV !== 'production') {
```
timfish: I'm not sure about this logic. When an app is packaged and run on a users machine, `NODE_ENV` will usually be undefined.

In the main process, you can detect if the app is packaged via:
```
process.defaultApp === undefined
```

Many will also want a way to not display the dialog in development as its 
a) too narrow to properly display stack traces
b) you can't copy and paste from the dialog

src/dispatch.ts
```diff
@@ -113,18 +129,15 @@ export class ElectronFrontend implements CommonFrontend {
...
-  }
+/**
+ * This either calls init on main with node interations or renderer
+ * with browser integrations.
+ * @param options Options
+ */
+export function specificInit(options: ElectronOptions): void {
+  process.type === 'browser'
+    ? module.require('./main').init(options)
+    : require('./renderer').init(options);
```
omeid: Isn't this logic backwards? :thinking:

HazAT: It's confusing I know https://electronjs.org/docs/api/process#processtype

PR #99, opened 2018-07-24
Feature/update core

HazAT:
- We update the scope  and persist it to disk before merging it.
- Updated SDK info.
- Expose `getIntegrations()` to return all used integrations

### DIFF
--- a/src/dispatch.ts
+++ b/src/dispatch.ts
@@ -1,6 +1,6 @@
 import { ClientClass, DSN } from '@sentry/core';
 import { getDefaultHub as getHub, Scope } from '@sentry/hub';
-import { Breadcrumb, SentryEvent, SentryResponse } from '@sentry/types';
+import { Breadcrumb, Integration, SentryEvent, SentryResponse } from '@sentry/types';
 import { CommonClient, ElectronOptions } from './common';
 
 // tslint:disable:no-var-requires no-unsafe-any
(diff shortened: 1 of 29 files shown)


src/main/integrations/sdkinformation.ts
```diff
@@ -0,0 +1,35 @@
...
+   * @inheritDoc
+   */
+  public name: string = 'SDKInformation';
+
+  /**
+   * @inheritDoc
+   */
+  public install(): void {
+    getDefaultHub().addEventProcessor(async (event: SentryEvent) => {
+      event.sdk = {
```
kamilogorek: No mutations plx

src/dispatch.ts
```diff
@@ -137,7 +105,16 @@ export class ElectronClient implements CommonClient {
...
  */
 export function specificInit(options: ElectronOptions): void {
-  process.type === 'browser'
-    ? module.require('./main').init(options)
-    : require('./renderer').init(options);
+  process.type === 'browser' ? module.require('./main').init(options) : require('./renderer').init(options);
+}
+
+/** Convenience interface used to expose Integrations */
+export interface Integrations {
```
kamilogorek: Isn't integrations type just `Integration[]`?

src/main/index.ts
```diff
@@ -4,20 +4,21 @@ import { initAndBind } from '@sentry/core';
...
-import { Electron } from './integrations/electron';
-import { OnUncaughtException } from './integrations/onuncaughtexception';
+import { Electron, OnUncaughtException, SDKInformation } from './integrations';
 import { NetTransport } from './transports/net';
 export { MainClient } from './client';
 export { MainBackend } from './backend';
+export { Integrations as NodeIntegrations } from '@sentry/node';
+
+// tslint:disable-next-line:variable-name
+export const ElectronIntegrations = { Electron, OnUncaughtException, SDKInformation };
```
kamilogorek: It's not consistent with what we have in `node` and `browser`. We export `Integrations = [...]` there. So an Array, not object.

PR #112, opened 2018-09-18
feat: Use 4.x.x release

HazAT:


package.json
```diff
@@ -30,52 +30,51 @@
...
   },
   "dependencies": {
-    "@sentry/browser": "4.0.0-rc.1",
-    "@sentry/core": "4.0.0-rc.1",
-    "@sentry/hub": "4.0.0-rc.1",
-    "@sentry/node": "4.0.0-rc.1",
-    "@sentry/minimal": "4.0.0-rc.1",
-    "@sentry/utils": "4.0.0-rc.1",
-    "@sentry/types": "4.0.0-rc.1",
+    "@sentry/browser": "4.0.0",
```
timfish: We could make these `^4.0.0` so that bug fixes in all the core stuff doesn't require bumping dependencies so often?

HazAT: Good point!

PR #121, opened 2018-09-28
Fixed Deep Dive URLs to point to the new locations

SecretObsession:
The README Deep Dive section URLs are all 404 now. This PR fixes those links to point to the new URLs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,10 +84,10 @@ Sentry.captureEvent({
 
 ## Deep Dive
 
-* [Configuration](https://docs.sentry.io/clients/electron/config/)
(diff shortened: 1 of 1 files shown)


jan-auer:
@HazAT I didn't realize we had moved these docs. The SDK is a strict super set of JavaScript just like React Native. We should either move RN, too, or move Electron out again (which I am slighly in favor of. Will keep this open until this is settled.

HazAT:
@jan-auer The plan is to move RN as soon we rewrote it to the new API.
That's why electron was moved already, we can merge this.

PR #122, opened 2018-09-29
Fix: Change Electron integration functionality 

timfish:
- Moves event normalisation out from the Electron integration
- Moves instrumented electron breadcrumbs into the Electron integration

This way, if a user sets `defaultIntegrations: false`, path normalisation will still work but the Electron breadcrumbs will no longer be written.

### DIFF
--- a/src/main/backend.ts
+++ b/src/main/backend.ts
@@ -2,8 +2,6 @@ import {
   app,
   crashReporter,
   ipcMain,
-  powerMonitor,
-  screen,
   // tslint:disable-next-line:no-implicit-dependencies
 } from 'electron';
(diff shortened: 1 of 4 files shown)


timfish:
The tests pass but I'm not 100% sure I've moved `normalizeEvent` to the right place

jan-auer:
@timfish could you please rebase?

timfish:
The AppVeyor tests have become rather flaky.

jan-auer:
Not sure what's going on with AppVeyor. Tests have succeeded but the build seems to get stuck repeatedly. Merging this now.

PR #134, opened 2018-10-23
fix: remove unnecessary requiring of package.json

### LINKED ISSUE #132: Critical dependency: the request of a dependency is an expression
I got the following error when importing `@sentry/electron` with webpack:

```
WARNING in ./node_modules/@sentry/electron/dist/main/context.js 79:11-31
Critical dependency: the request of a dependency is an expression
 @ ./node_modules/@sentry/electron/dist/main/client.js
 @ ./node_modules/@sentry/electron/dist/main/index.js
 @ ./node_modules/@sentry/electron/dist/dispatch.js
 @ ./node_modules/@sentry/electron/dist/index.js
```

pmccarren:
There is not a need to load the `package.json`, as electron
remote.app provides us this functionality natively. We were
loading `package.json` to parse the package name, which is what
`getAppName()` already returns (and we already use this too).

Ref: https://electronjs.org/docs/api/app#appgetname

`getPackageJson()` was only being used by `getEventDefaults()`.
It is no longer needed as we already have all required `package.json`
data to build the event defaults.

**Note:** _(Ulterior motive):_
This has the benefit of removing an expression based `require()`,
which makes webpack happy! 🎉 

Closes #132

### DIFF
--- a/src/main/context.ts
+++ b/src/main/context.ts
@@ -90,12 +90,6 @@ function matchFirst(regex: RegExp, text: string): string | undefined {
   return match ? match[1] : undefined;
 }
 
-/** Synchronously loads this app's package.json or throws if not possible. */
-function getPackageJson(): PackageJson {
-  const packagePath = join(app.getAppPath(), 'package.json');
-  return require(packagePath) as PackageJson;
-}
-
 /** Returns the build type of this app, if possible. */
 function getBuildType(): string | undefined {
   if (process.mas) {
@@ -253,7 +247,7 @@ async function getEventDefaults(): Promise<SentryEvent> {
     },
     environment: process.defaultApp ? 'development' : 'production',
     extra: { crashed_process: 'browser' },
-    release: `${getPackageJson().name}@${app.getVersion()}`,
+    release: `${app.getName()}@${app.getVersion()}`,
     user: { ip_address: '{{auto}}' },
   };
 }
--- a/src/main/context.ts
+++ b/src/main/context.ts
@@ -23,11 +23,6 @@ interface OsContext {
   kernel_version?: string;
 }
 
-/** Shim interface to access this app's package.json. */
-interface PackageJson {
-  name: string;
-}
-
 /** Linux version file to check for a distribution. */
 interface DistroFile {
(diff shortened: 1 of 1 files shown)


jan-auer:
Hi @pmccarren and thanks for putting together a PR. We explicitly required package.json over reading `app.getName()` for two reasons:

 1. It prefers the `productName` key, as documented [here](https://github.com/electron/electron/blob/master/docs/api/app.md#appgetname)
 2. You can override this value with `app.setName()`.

However, we require this string to be a valid package identifier, so no spaces are allowed, among other things. When implemented originally, there was no other stable way to retrieve the original package name.

We could however trick webpack into leaving this require, e.g. by using `module.require` instead. This should also solve #132.

I'm surprised this hasn't come up before for some of our other users who are using webpack for bundling the main process code. Did you do special configuration or is this also happening with, say, a boilerplate code?

pmccarren:
Hi @jan-auer thanks for the info! I wasn't aware of the requirement of a valid package-identifier. Definitely see the appeal of reading in the package.json.

Your trick with `module.require` works like a charm! I'm not familiar with the nuances of `require` v. `module.require`, but a quick test didn't reveal any issues. Any reason not to use `module.require`? Happy to submit a PR for the change (the simple one-liner), unless it is something you'd like to handle on your end?

I'm using the [electron-webpack](https://github.com/electron-userland/electron-webpack) boilerplate, which itself does plenty of webpack magic. Can't speak to what is being used in #132, but perhaps webpack gets grumpy with any expression based require call?

jan-auer:
I just learned out that webpack 4.16.2 [changed this](https://github.com/webpack/webpack/issues/7728) and now also detects `module.require`. We will fix this (cc @HazAT) in the remaining places where we use this hack. 

Since this code is only executed on the main process, you can use `global.module.require` for now and we'll make sure to find a more sustainable workaround in future. The only reason not to use it is that it might break in future if webpack starts supporting `global.module.require` as well, but I guess we could live with that.

PR definitely welcome, and thank you for taking the time!

pmccarren:
@jan-auer I just opened #136 - closing this PR to move to the new approach.

PR #136, opened 2018-10-24
fix: use module.require in place of require for dynamic includes

pmccarren:
This is the proposed alternative solution of PR #134 

Looking though the code it looks like the `require()` call in `getPackageJson` is the only one that needs the `module.require` hack to be introduced for. If there are any that I missed, please let me know!

Ref: webpack/webpack#7728
Closes: getsentry/sentry-electron#132


HazAT:
I will just go ahead an merge it, thanks for the PR, I will change it in my code and release it.

pmccarren:
No problem, glad to help!

PR #153, opened 2019-02-01
escape base path in normalizeUrl function

marcello3d:
This PR handles app paths that have parentheses and other RegExp characters.


timfish:
Thanks for this!

It looks like there are some linting failures in CI. Could you try running `yarn fix` to see if that fixes it?

PR #154, opened 2019-02-01
Test case for broken stack trace when path has parenthesis

marcello3d:
When our electron app has a parenthesis in the name, we get bad stack traces that don't SourceMap properly.

I was able to fix part of the problem with https://github.com/getsentry/sentry-electron/pull/153 but that only helps with browser errors.

Errors in the main process seem to be parsed elsewhere (maybe by the `stack-trace` module?).

In any case, this PR doesn't solve the bug, just provides a test for it.


timfish:
From [this quick test](https://regex101.com/r/C4PuTR/1), it looks like this might be fixable in `stack-trace` by making the first capturing group of the regex less greedy.

So [this line](https://github.com/felixge/node-stack-trace/blob/4fd379ee78965ce7ce8820b436f1b1b590d5dbcf/lib/stack-trace.js#L42):
```typescript
var lineMatch = line.match(/at (?:(.+)\s+\()?(?:(.+?):(\d+)(?::(\d+))?|([^)]+))\)?/);
// would become
var lineMatch = line.match(/at (?:(.+?)\s+\()?(?:(.+?):(\d+)(?::(\d+))?|([^)]+))\)?/);
```

marcello3d:
That's what I found as well, but when I tried to write tests on the repo against the bug I wasn't able to get it working. :-(

I'm also not sure if there are regressions with that change.

timfish:
The tests appear to pass with the change but there aren't that many tests and no tests for Windows paths. It also has 3m downloads per week and no release since 2017 so it seems like a risky change.

If I have time over the weekend, I might try submitting a PR (or two) adding more tests and the above change and see how it goes down 🤷‍♂️

HazAT:
This will be fixed one we move to `5.x` of the `@sentry/core` SDKs

HazAT:
I will close this since we released new major bump with using 5.x SDKs

timfish:
@HazAT I think this PR was a test case to ensure that the `stack-trace` was fixed so it should have been merged after the v5 update.

HazAT:
OK so good to merge now, lol :)

Issue #142, opened 2018-11-09
TypeError: mod.require is not a function


Djiit:
Hello there,

Using electron 3 and sentry-electron 0.13.0, I can't initialise Sentry and get this error when the app starts : 

```
TypeError: mod.require is not a function
    at Object.dynamicRequire (webpack:///./node_modules/@sentry/utils/misc.js?:10:16)
    at Object.specificInit (webpack:///./node_modules/@sentry/electron/dist/dispatch.js?:129:41)
    at init (webpack:///./node_modules/@sentry/electron/dist/sdk.js?:52:16)
    at eval (webpack:///./src/sentry.js?:5:3)
    at Object../src/sentry.js (/Users/jtanay/Projects/misc/desktop-app/dist_electron/index.js:3730:1)
    at __webpack_require__ (/Users/jtanay/Projects/misc/desktop-app/dist_electron/index.js:20:30)
    at eval (webpack:///./src/background.js?:9:65)
    at Module../src/background.js (/Users/jtanay/Projects/misc/desktop-app/dist_electron/index.js:3719:1)
    at __webpack_require__ (/Users/jtanay/Projects/misc/desktop-app/dist_electron/index.js:20:30)
    at eval (webpack:///multi_./src/background.js?:1:18)
```

I just copy-pasted the docs from here : https://docs.sentry.io/error-reporting/quickstart/?platform=electron#configure-the-sdk and currently init Sentry only in my background process.

Did I miss something ?

Thanks for you help

Superjo149:
Same issue here, I switched back to node & browser for now. This issue also happened in the earlier versions as well. #92

alxmiron:
Also got this one

rhyek:
I am also getting this when calling `init` in the main process with electron 2.0.9.

timfish:
This is almost certainly Webpack related but our config is working fine. If you can supply a minimal reproduction of the issue I might be able to point you in the right direction...

Norbz:
Getting the same error here, using Electron 3
We are using the Vue-CLI boilerplate, so you should be able to reproduce the error easily using https://github.com/nklayman/vue-cli-plugin-electron-builder

Not really sure were to start digging thought, first time using Sentry

koalalorenzo:
Same error here. I am not sure how to work with it or having a workaround for Webpack.
Any solution?

Norbz:
Hi there,

Any progress on this ? This is really a shame we can't use sentry because of a compilation error...

[...]

PR #180, opened 2019-08-23
feat: Use v5 of JS SDKs

HazAT:
### Tests

- [x] `1.7.16`
- [x] `1.8.8`
- [x] `2.0.10`
- [x] `3.0.2`
- [x] `4.2.10`
- [x] `5.0.10`
- [x] `6.0.7`


claygriffiths:
Just chiming in to say we'd love to see this happen!

HazAT:
@claygriffiths Cool :) 
So, I couldn't make the test run on electron 5/6 didn't spend enough time with it.
I tested it manually though and it seems to work just fine.
If you have time testing this branch in your app it would bring more assurance to ship this regardless.

HazAT:
It's ready to review.
Test look ok'ish, some are failing on Windows we will investigate why.

I would love to have some app that already use the SDK to upgrade cc @timfish @claygriffiths

timfish:
@HazAT how easy is it to get your build tooling publishing branches like this to some sort of "alpha" tag on npm? It would certainly help speed up testing.

Currently to test it with our apps, on all platforms we support I have to `git clone/checkout` this branch, `yarn build`, `yarn link`, `yarn link @sentry/electron` and then build our app. If it was published on npm I could update the dependency in single place and let our CI get on with all the work.

Using yarn link also doesn't resolve common dependencies that an app might have with `@sentry/electron` which means it's not strictly testing how it will function later when the dependency comes straight from npm.

timfish:
Testing:
- [x] Windows
  - [x] JS
  - [x] Minidump
- [ ] macOS
  - [ ] JS
  - [ ] Minidump
- [ ] Linux
  - [ ] JS
  - [ ] Minidump

timfish:
We are currently [doing this](https://github.com/getsentry/sentry-electron/issues/142#issuecomment-469286795) to get around the issues with Webpack bundling. 

Building our app works find on Windows but on macOS Catalina it fails when used via `yarn link` with an error:

```
    ERROR in /Users/timfish/Documents/Repositories/sentry-electron/dist/renderer/backend.js
    Module not found: Error: Can't resolve '@sentry/browser/dist/backend' in '/Users/timfish/Documents/Repositories/sentry-electron/dist/renderer'
     @ /Users/timfish/Documents/Repositories/sentry-electron/dist/renderer/backend.js 4:18-57
     @ /Users/timfish/Documents/Repositories/sentry-electron/dist/renderer/index.js
     @ ./src/error-reporting.ts
// snip
```

`@sentry/browser/dist/backend` is there if I follow normal node resolution paths but I've yet to work out if this is caused by `yarn link`, Webpack or maybe even [this](https://twitter.com/MarshallOfSound/status/1136526239139635200)... or a combination of all three!

HazAT:
@timfish Good point, will release beta package.

HazAT:
@timfish `1.0.0-beta.0` should be on npm also tagged with `next`.
cc @claygriffiths

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 There shouldn't be any breaking changes, interally we needed to do some changes because we were upgrading to use `5.x`
 of the JS/Node SDK.
 
+**Breaking**: We persist scope data to disk in case of a native crash. Since the internal scope data changed, existing
```
HazAT: cc @timfish

claygriffiths:
Thanks all who were involved in implementing/testing this! We're looking forward to upgrading 😃

PR #194, opened 2019-11-15
Fix: Electron v6-7

### LINKED ISSUE #144: Catch file system failures
We're seeing a couple of failures while handling Electron dump artifacts. I'm guessing maybe over zealous virus scanner:
```
Error: EBUSY: resource busy or locked, open 'C:\Users\username\AppData\Local\Temp\OurApp Crashes\01692099-317b-4ea7-b6c1-3d5fbee3c3aa.dmp'
```

```
Error: ENOENT: no such file or directory, unlink 'C:\Users\username\AppData\Local\Temp\OurApp Crashes\82e067bc-c36e-4093-98c6-d303ee6e08f9.txt'
```

### LINKED ISSUE #176: Setting Origin on request headers causes sentry-electron to crash Electron
Many APIs will reject requests without an `Origin` header, or an Origin set to `file://*`. To mitigate this in Electron, [it's common](https://github.com/electron/electron/issues/6859) to add a dummy Origin header like this:
```typescript
app.on('ready', () => {
  session.defaultSession.webRequest.onBeforeSendHeaders(
    (details: any, callback: any) => {
      details.requestHeaders.Origin = 'www.some-random-origin.com';
      callback({ cancel: false, requestHeaders: details.requestHeaders });
    }
  );
});
```

With the above, `sentry-electron@0.17.1` will immediately crash on attempting to upload a minidump which will usually be during startup.. 

This regression is no doubt due to a recent change to `electron-fetch`:
> v1.2.0
> - Using electron's `defaultSession` by default

### LINKED ISSUE #192: Electron 7 crashes on Sentry.init
Updated Electron to 7.1.0 and since then the app crashes the moment `Sentry.init` is called. Does Sentry support that version?

Thanks

timfish:
- Add issue template
- Update dependencies
- Fix linting issues
- Update and add all Electron versions for e2e tests
- Stop allowing test failures on Windows
- Use single instance of TestServer for all tests
- Use crashpad logic for Electron >= v6 on Windows
- Use `node-fetch` rather than `electron-fetch`
- Use `win-ca` to load Windows root CA's to give same behaviour as `electron-fetch`
- Use Sentry logger to trace some caught exceptions

Fixes #192, Fixes #176 and Fixes #144


timfish:
No idea why the Windows tests are not passing. The output suggests its completed correctly:

image

src/common.ts
```diff
@@ -64,7 +64,7 @@ export interface CommonClient extends Client<ElectronOptions> {
   captureMinidump(path: string, event?: Event, scope?: Scope): string | undefined;
 
   /**
-   * @inheritdoc {@link BrowserClient.showReportDialog}
+   * Shows Report Dialog}
```
HazAT:
```suggestion
   * Shows Report Dialog
```

PR #203, opened 2019-12-20
Feat: Allow usage without remote module

timfish:
cc @koenoe

This allows usage of `sentry-electron` with the `remote` module disabled in renderers:
```
const mainWindow = new BrowserWindow({
  webPreferences: {
    enableRemoteModule: false
  }
})
```
To support this you have to pass the app name to sentry options:
```
const { init } = require('@sentry/electron');

init({
  dsn: 'https://343434343434343@sentry.io/5656565',
  appName: 'My Great App',
});
```

### TODO
Add some tests!

### DIFF
--- a/src/main/backend.ts
+++ b/src/main/backend.ts
@@ -13,7 +13,7 @@ import { Dsn, forget, logger, SentryError } from '@sentry/utils';
 import { app, crashReporter, ipcMain } from 'electron';
 import { join } from 'path';
 
-import { CommonBackend, ElectronOptions, getName, IPC_EVENT, IPC_PING, IPC_SCOPE } from '../common';
+import { CommonBackend, ElectronOptions, getNameFallback, IPC_EVENT, IPC_PING, IPC_SCOPE } from '../common';
 import { captureMinidump } from '../sdk';
 
 import { normalizeUrl } from './normalize';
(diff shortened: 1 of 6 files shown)


timfish:
@HazAT the build is failing for the same reason as your `feat/allow-no-remote` branch. `Dsn` has moved from `utils` to `core` but isn't exported?

src/main/backend.ts
```diff
@@ -4,16 +4,17 @@ import {
   captureEvent,
   captureMessage,
   configureScope,
+  Dsn,
```
HazAT:
```suggestion
```
`Dsn` lives in `@sentry/utils` now.

src/main/uploader.ts
```diff
@@ -1,5 +1,6 @@
+import { Dsn } from '@sentry/core';
```
HazAT:
```suggestion
```
Same here

src/main/uploader.ts
```diff
@@ -1,5 +1,6 @@
...
 import { Event } from '@sentry/types';
-import { Dsn, logger } from '@sentry/utils';
+import { logger } from '@sentry/utils';
```
HazAT:
```suggestion
import { Dsn, logger } from '@sentry/utils';
```

src/main/backend.ts
```diff
@@ -4,16 +4,17 @@ import {
...
   getCurrentHub,
   Scope,
 } from '@sentry/core';
 import { NodeBackend } from '@sentry/node/dist/backend';
 import { Event, EventHint, Severity, Transport, TransportOptions } from '@sentry/types';
-import { Dsn, forget, logger, SentryError } from '@sentry/utils';
+import { forget, logger, SentryError } from '@sentry/utils';
```
HazAT:
```suggestion
import { Dsn, forget, logger, SentryError } from '@sentry/utils';
```

Issue #195, opened 2019-11-15
user data missing on uncaught exception


crussell52:
sdk: sentry-electron 1.0
platform: mac-os, linux (appImage)

# Description

After setting `user`.id and a tag using `configureScope()`, unhandled exception does not report user info, but reports tag.

To distill, this, I turned off all integration and inspected scope during `Process.unhandledException` event and found user info absent.

# Repro

```js
const {app} = require('electron');
const Sentry = require('@sentry/electron');

Sentry.init({
    dsn: myDSN,
    debug: true,
    defaultIntegrations: false
});

Sentry.configureScope((scope) => {
    scope.setUser({id: 'johndoe'});
    scope.setTag("foo", "bar");
    console.log('configureScope', scope._user, scope._tags); // OK!
});

Sentry.withScope(scope => {
    console.log('withScope', scope._user, scope._tags); // OK !!
});

process.on('uncaughtException', (e) => {
    Sentry.withScope((scope) => {
        console.log('uncaughtException', scope._user, scope._tags); // USER gone!
        app.exit();
    })
});

setTimeout(() => junk(), 10000);
```

crussell52:
While working with #196 , I found some extra clues:
  - During `uncaughtException` handling, the on-disk scope (`userData/sentry/scope_v2.json`) appears to be used instead of the in-memory scope.
  - The on-disk scope does not include user data set via `configureScope()` but it does include some version of the tag data set at the same time.

crussell52:
Any progress on this. Can't move forward with sentry as a solution with this pending.
:(

HazAT:
I hope I find some time to take a look at this, we gladly accept PRs fixing this if you find time yourself.

Thanks for reporting!

crussell52:
@HazAT Have you found time to circle back to this?  

I opened PR #210 with an e2e test that checks for user data. The test obviously fails. I _believe_ the test is valid, but I could be wrong.

jjustin3:
[Here is the problematic line](https://github.com/getsentry/sentry-electron/blob/master/src/main/backend.ts#L68).

When laying the code out like this
```
const { init } = require('@sentry/electron/dist/main');
const Sentry = require('@sentry/electron');

init({ ... });

Sentry.setUser({ ... }); // or even Sentry.configureScope(scope => { ... })
```
[...]

PR #210, opened 2020-01-10
Add e2e test for testing user data w/ exception

crussell52:
Test to demonstrate #195


crussell52:
This test assumes that `testServer.events[0].event.data.user` is the correct place to look for user data, but that is mostly a guess since this seems to be the first e2e test that looks for user data.

HazAT:
Thx!

HazAT:
Thanks for the investigation, superseded by https://github.com/getsentry/sentry-electron/pull/242

PR #215, opened 2020-02-11
Fix: IPC objects valid for structured clone algorithm

### LINKED ISSUE #214: Sentry is sending non-cloneable JavaScript objects through IPC
Electron 8 started throwing deprecation warnings for IPC method calls that have non-cloneable JavaScript objects as params. This is so they can move to a more efficient serializer from the V8 engine.

You can check more details in here:
https://github.com/electron/electron/pull/20214

From my investigations, Sentry is sending such non-cloneable objects at the following lines:
https://github.com/getsentry/sentry-electron/blob/master/src/renderer/backend.ts#L51
https://github.com/getsentry/sentry-electron/blob/master/src/renderer/backend.ts#L61

It still works on Electron 8, with deprecation warnings, and will stop working on Electron 9 according to their roadmap.

Personal note: I HATE when I have deprecation warnings in my application.

Thanks!

timfish:
Fixes #214

### DIFF
--- a/src/renderer/backend.ts
+++ b/src/renderer/backend.ts
@@ -48,7 +48,9 @@ export class RendererBackend extends BaseBackend<ElectronOptions> implements Com
    * @inheritDoc
    */
   public sendEvent(event: Event): void {
-    ipcRenderer.send(IPC_EVENT, event);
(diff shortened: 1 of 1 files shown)


shahkeyur:
Any updates here ?

PR #236, opened 2020-07-03
ci: Try GH Actions

HazAT:


.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
+name: 'Test'
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '*'
```
flub, line 8: I think you can save two lines by leaving this out as is done in https://github.com/getsentry/symbolicator/pull/235/files. This way is pretty readable though, so could be preferable.

PR #240, opened 2020-07-07
fix: Create error object in case we cant detect it

HazAT:
This fixes a weird bug where we capture an exception which isn't detected as an error and therefore be captured as `Non-Error exception captured with keys: message, name, stack`

### DIFF
--- a/src/main/integrations/onuncaughtexception.ts
+++ b/src/main/integrations/onuncaughtexception.ts
@@ -1,6 +1,7 @@
 import { getCurrentHub, NodeClient } from '@sentry/node';
 import { Event, Integration, Severity } from '@sentry/types';
 import { dialog } from 'electron';
+import { isError } from '@sentry/utils';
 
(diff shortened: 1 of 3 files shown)


src/main/integrations/onuncaughtexception.ts
```diff
@@ -37,8 +38,16 @@ export class OnUncaughtException implements Integration {
             level: Severity.Fatal,
           }));
 
+          let theError = error;
+          if (!isError(error) && error.message && error.stack && error.name) {
```
jan-auer, line 42: Note that this is falsy if any of the attributes is empty. Do you rather mean to check for `undefined`?

HazAT: Good point, I made it to only check for `stack` which makes it less strict.

jan-auer: Sounds good!

Issue #246, opened 2020-07-14
'crashed' event is deprecated as of Electron 8.4


jleongDialpad:
## Versions + Platform

- [ ] SDK version - `@sentry/electron@v1.3.0`
- [ ] Electron version - `electron@v8.4.0`
- [ ] Platform - `Windows/macOS/Linux`

## Description

Electron 8.4.0 has deprecated the 'crashed' event in favour of 'render-process-gone'. The new event gives finer detail over why the render process disappeared
```
Added new render-process-gone event on app to replace the renderer-process-crashed event. #24314
Added new render-process-gone event to replace the crashed event. #24308
```

https://www.electronjs.org/releases/stable?version=8#release-notes-for-v840

Notes of deprecation:
https://www.electronjs.org/docs/api/web-contents#event-crashed-deprecated
> This event is superceded by the render-process-gone event which contains more information about why the render process dissapeared. It isn't always because it crashed. The killed boolean can be replaced by checking reason === 'killed' when you switch to that event.

Current sentry-electron version handle the event here:
https://github.com/getsentry/sentry-electron/blob/d20053d89391c56dabcfba07e349f1c48466f90c/src/main/backend.ts#L216

To support older and present versions the proposed update would be to handle both crashed and render-process-gone event here.

HazAT:
We will fix this in the next version :)

PR #247, opened 2020-07-15
feat: render-process-gone

HazAT:
Fixes https://github.com/getsentry/sentry-electron/issues/246

Adds electron context with crash reason
CleanShot 2020-07-15 at 11 53 06

### DIFF
--- a/src/main/backend.ts
+++ b/src/main/backend.ts
@@ -204,38 +204,34 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
       forget(this._sendNativeCrashes({}));
     }
 
+    /**
+     * Helper function for sending renderer crashes
+     */
+    const sendRendererCrash = async (contents: Electron.WebContents, details?: Electron.Details) => {
+      try {
+        await this._sendNativeCrashes(this._getRendererElectronContext(contents, details));
+      } catch (e) {
+        console.error(e);
+      }
+
+      addBreadcrumb({
+        category: 'exception',
+        level: Severity.Critical,
+        message: 'Renderer Crashed',
+      });
+    };
+
     // Every time a subprocess or renderer crashes, start sending minidumps
     // right away.
     app.on('web-contents-created', (_, contents) => {
-      if (getElectronVersion().major > 8 || (getElectronVersion().major === 8 && getElectronVersion().minor >= 4)) {
+      // We need to update this once this lands in 9.x
+      if (getElectronVersion().major === 8 && getElectronVersion().minor >= 4) {
         contents.on('render-process-gone', async (_event, details) => {
-          try {
-            await this._sendNativeCrashes(this._getRendererElectronContext(contents, details));
-          } catch (e) {
-            console.error(e);
-          }
-
-          addBreadcrumb({
-            category: 'exception',
-            level: Severity.Critical,
-            message: 'Renderer Crashed',
-            timestamp: new Date().getTime() / 1000,
-          });
+          await sendRendererCrash(contents, details);
         });
       } else {
         contents.on('crashed', async () => {
-          try {
-            await this._sendNativeCrashes(this._getRendererElectronContext(contents));
-          } catch (e) {
-            console.error(e);
-          }
-
-          addBreadcrumb({
-            category: 'exception',
-            level: Severity.Critical,
-            message: 'Renderer Crashed',
-            timestamp: new Date().getTime() / 1000,
-          });
+          await sendRendererCrash(contents);
         });
       }
 
--- a/src/main/backend.ts
+++ b/src/main/backend.ts
@@ -20,6 +20,7 @@ import { normalizeUrl } from './normalize';
 import { Store } from './store';
 import { NetTransport } from './transports/net';
 import { MinidumpUploader } from './uploader';
+import { getElectronVersion } from './utils';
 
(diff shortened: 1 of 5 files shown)


src/main/utils.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Returns the major version of electron
+ */
+export function getElectronVersion(): { major: number; minor: number; patch: number } {
+  const versions = process.versions.electron.split('.').map(val => parseInt(val, 10));
```
timfish, line 5: This will not work for all currently used semver strings (ie. `v10.0.0-beta.11`)

This should be more reliable:
```typescript
const { groups } = /(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)/.exec(process.versions.electron);
return groups;
```

src/main/utils.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Returns the major version of electron
+ */
+export function getElectronVersion(): { major: number; minor: number; patch: number } {
```
AbhiPrasad, line 4: We should cache this result right? I know it’s not that expensive, but it’s reasonable to assume that it won’t change in application lifetime + it current route is hot code path right? Maybe this is over optimization lol

src/main/backend.ts
```diff
@@ -203,23 +204,36 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
...
-          console.error(e);
-        }
-
-        addBreadcrumb({
-          category: 'exception',
-          level: Severity.Critical,
-          message: 'Renderer Crashed',
-          timestamp: new Date().getTime() / 1000,
+      // We need to update this once this lands in 9.x
+      if (getElectronVersion().major === 8 && getElectronVersion().minor >= 4) {
```
timfish, line 228: We also need to support `>= v10.x.y`.

If it's safe to assume that the next 9.x release will be 9.2 and will include the backport, we need:

```
getElectronVersion().major === 8 && getElectronVersion().minor >= 4 || 
getElectronVersion().major === 9 && getElectronVersion().minor >= 2 ||
getElectronVersion().major >= 10
```

src/main/backend.ts
```diff
@@ -316,12 +335,20 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
   }
 
   /** Returns extra information from a renderer's web contents. */
-  private _getRendererExtra(contents: Electron.WebContents): { [key: string]: any } {
+  private _getRendererElectronContext(contents: Electron.WebContents, details?: Electron.Details): Event {
     const customName = this._options.getRendererName && this._options.getRendererName(contents);
-
-    return {
+    const electronContext: { [key: string]: any } = {
```
AbhiPrasad, line 340:
```suggestion
    const electronContext: Record<string, any> = {
```

we can’t type this stronger? We are explicitly setting fields below.

src/main/backend.ts
```diff
@@ -316,12 +335,20 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
   }
 
   /** Returns extra information from a renderer's web contents. */
-  private _getRendererExtra(contents: Electron.WebContents): { [key: string]: any } {
+  private _getRendererElectronContext(contents: Electron.WebContents, details?: Electron.Details): Event {
```
AbhiPrasad, line 338: This returns the event, so I’m not sure `getElectronContext` is accurate here.

src/main/backend.ts
timfish: Or you could use `semver` and do:

```typescript
semver.satisfies(process.versions.electron, '>= 10.x || >=9.2 || >= 8.4')
```

HazAT: Better no new deps :|
But will change

timfish:
Ah also, I'm guessing many of the `reason`s don't actually result in a crash dump.

I'd guess `clean-exit`, `abnormal-exit`, `integrity-failure` and `launch-failure` are not going to result in a crash dump but users still might want events in Sentry for some of them.

No users are going to want to know about `clean-exit` and in many cases users won't want to report `abnormal-exit` either if they're using non-zero exit codes when their app is called form the command line.

image

src/main/utils.ts
HazAT: Thought the same, but `process.versions.electron` is a readonly string, a "cache" would do the same.

timfish:
Actually, `abnormal-exit` is the exit code from the renderer so this would not be used in cli apps.

The different `reason`s need handling something like this:

| `reason` | report to Sentry | crash dump? |
| - | - | - |
| `clean-exit` | ❌ | ❌ |
| `abnormal-exit`<sup>1</sup> | ✔ | ❌ |
| `killed` | ❔<sup>2</sup> | ❌ |
| `crashed` | ✔ | ✔ |
| `oom` | ✔ | ✔ |
| `launch-failure` | ✔ | ❔<sup>3</sup> |
| `integrity-failure` | ✔ | ❔<sup>3</sup> |

1. It would be great to be able to log the abnormal exit code
2. Do users want to get a report if their app was killed?
3. No idea if crash dumps are created for these

HazAT:
I am going to merge this, let's improve it further.
I am working on providing an ESM build which has some internal refactoring which will be hard to merge.

xiaoxiangmoe:
@HazAT When will this be released?

PR #252, opened 2020-07-24
feat: Remove deps and use Envelope Endpoint

HazAT:
This greatly reduced bundle size and make the minidump upload a bit more robust.

Size of bundle before:
image

Size after:
image

This will become a major version bump because it will only work with on-premise Sentry `20.6.0` or older since we use the envelope endpoint. https://develop.sentry.dev/sdk/envelopes/


src/main/uploader.ts
```diff
@@ -95,6 +82,35 @@ export class MinidumpUploader {
...
+    const envelopeHeaders = JSON.stringify({
+      event_id: event.event_id,
+      sent_at: new Date(timestampWithMs() * 1000).toISOString(),
+    });
+    const itemHeaders = JSON.stringify({
+      content_type: 'application/json',
+      type: 'event',
+    });
+
+    const stat = await statAsync(minidumpPath);
```
jan-auer, line 98: You can avoid this stat, since you can simply query for `minidumpContent.length` below. That will return the binary size of the buffer. That way, you avoid differences between the FS metadata and the file contents, if any.

src/main/uploader.ts
```diff
@@ -103,23 +119,27 @@ export class MinidumpUploader {
...
-      const body = new FormData();
-      body.append('upload_file_minidump', fs.createReadStream(request.path));
-      body.append('sentry', JSON.stringify(request.event));
-      const response = await fetch(this._url, { method: 'POST', body });
+      const requestForTransport = await this._toMinidumpRequest(request.event, request.path);
+      const response = await transport.sendRequest(requestForTransport);
 
       // Too many requests, so we queue the event and send it later
-      if (response.status === CODE_RETRY) {
+      if (response.status === Status.RateLimit) {
```
jan-auer, line 135: Here, we have to make two changes:
1. On a `429` response, we should never retry. 
2. On top of that, the envelope endpoint may return a `429`, even if the event was accepted. Please check the corresponding `X-Sentry-Rate-Limits` header if a rate limit for the `"error"` category is supported, and ignore the ones for attachments for now.

We can later follow up with a more nuanced rate limiting logic.

src/main/uploader.ts
```diff
@@ -95,6 +82,35 @@ export class MinidumpUploader {
     }/api/${projectId}/minidump?sentry_key=${user}`;
   }
 
+  /**
+   * Create minidump request to dispatch to the transpoirt
+   */
+  private async _toMinidumpRequest(event: Event, minidumpPath: string): Promise<SentryElectronRequest> {
+    const envelopeHeaders = JSON.stringify({
+      event_id: event.event_id,
+      sent_at: new Date(timestampWithMs() * 1000).toISOString(),
```
jan-auer, line 91: It's probably worth noting that every time you retry a request, you have to generate a new `sent_at` timestamp. Otherwise, clock drift correction will see an old timestamp and assume your clock is lagging behind.

It seems that you do this correctly, can you confirm? I'd suggest to add a comment at some appropriate place for the future.

src/main/transports/net.ts
```diff
@@ -24,6 +31,26 @@ export class NetTransport extends Transports.BaseTransport {
    * @inheritDoc
    */
   public async sendEvent(event: Event): Promise<Response> {
+    const envelopeHeaders = JSON.stringify({
+      event_id: event.event_id,
+      sent_at: new Date(timestampWithMs() * 1000).toISOString(),
```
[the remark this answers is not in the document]
HazAT: `timestampWithMs` has some internal logic how to get the "initial" clock reading. We use this clock everywhere so I would keep it like that for now.

PR #256, opened 2020-08-12
build: Switch @sentry/electron to eslint

AbhiPrasad:
🚀 

We should also think about adding dependabot so we can bump sentry-js versions.


HazAT:
🥇

PR #269, opened 2020-10-22
fix(uploader): Add slash to minidump upload URL

jan-auer:
The initial version of the Electron SDK used to send minidumps to an endpoint without the trailing slash since the server initially did not support this by mistake. This has long been fixed, and the current version of the SDK even requires the Envelope endpoint for most of its operation.

Both in the [endpoint docs](https://docs.sentry.io/platforms/native/guides/minidumps/#minidump-integration) and [Relay's operating guidelines](https://docs.sentry.io/product/relay/operating-guidelines/#request-routing), we only document the URL with a trailing slash.


HazAT (approved):

HazAT:
Thx

PR #271, opened 2020-10-29
fix: Add guard for phantom event

HazAT:


src/main/backend.ts
```diff
@@ -86,6 +86,13 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
    * @inheritDoc
    */
   public sendEvent(event: Event): void {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    if ((event as any).event && (event as any).event.__INTERNAL_MINIDUMP) {
+      // We are not 100% sure when such an event is happening, regardsless we def don't want to send it
```
jan-auer, line 91:
```suggestion
      // We are not 100% sure when such an event is happening, regardless we def don't want to send it
```

If there's some sort of debug mode, it would be great to print the stack trace or some other information that allows us to trace it back to its origin.

jan-auer (approved):
So I have a theory about this. Here, we call the generic `captureEvent`:

https://github.com/getsentry/sentry-electron/blob/f4776e730d8636b907453e957ef45f56a686a88a/src/main/backend.ts#L149-L150

This is a scope listener on a hub, so I would expect that it always routes to the right backend, but what if a different client is bound? Particularly, it says `crashed_process: browser` in those events, so potentially this is being routed through the wrong client?

I was wondering if we should just be calling the backend's `sendEvent` directly rather than the generic `captureEvent` there.

HazAT:
It has to go through `captureEvent` to have all the context for the event. I am going to ship this and lets see.

PR #272, opened 2020-11-03
updated dependencies

clawdaddy:
updated all dependencies (but not devDependencies) to get all of the sentry packages current. also changed a couple of things in the code itself to fit with new types from those packages


HazAT:
Thx, this already in master

Issue #249, opened 2020-07-21
BeforeSend not tirgger when process crash


callmeguaishu:
When I use the official example demo, 'BeforeSend' will be trigger when throw an error, but not working when process crash. Is it correct ?

timfish:
If you have a native crash in the main process, the process terminates and no more code is executed so this would be expected.

If you have a native crash in a renderer process, the main process should continue and `beforeSend` should be called.

callmeguaishu:
y

> If you have a native crash in the main process, the process terminates and no more code is executed so this would be expected.
> 
> If you have a native crash in a renderer process, the main process should continue and `beforeSend` should be called.

I crete a snetry instance in main process and another sentry instance in a renderer process, when renderer process crash, beforeSend(sentry instance in renderer process) not working, you mean that time, main process beforeSend will be called?

callmeguaishu:
![image](https://user-images.githubusercontent.com/22768593/89882482-7ab14900-dbf9-11ea-9fec-4f96c1938cf0.png)

just like this, I use office example, add beforeSend function in sentry.js to print event hint, and click renderer process crash, main terminal didn't show event hint, just find mini dump and upload. click main error or renderer error is in line with expectation

vyunikov:
The same is true for `addGlobalEventProcessor`, handler passed to it doesn't seem to be called if process crashes, though it works in cases when exception is thrown. Based on SDK docs, this handler is expected to get called in case of the crash either:
https://docs.sentry.io/platforms/javascript/guides/electron/enriching-events/attachments/#uploading-attachments

theicfire:
I've run into the same issue. I think what @timfish mentioned is not correct. When the electron process restarts, it looks for any minidump files to send, and then sends them. But `beforeSend` isn't called.

It looks like the code is doing this on purpose in `captureMinidump`, but I don't know why:
```
    // We are not calling _processEvent here since we do not have beforeSend for minidump crashes
    this._prepareEvent(event, scope, undefined).then(
```

Note that in `sentry-javascript`, `_processEvent` is where `beforeSend` is called. So it looks like this is skipped.

Is there a reason why `beforeSend` is being skipped in `captureMinidump`?

timfish:
Fixed in v3

PR #273, opened 2020-11-04
ref: Don't call beforeSend for internal minidump

HazAT:


src/main/backend.ts
```diff
@@ -154,8 +147,17 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
...
-          captureEvent({ __INTERNAL_MINIDUMP: true });
+          getCurrentHub().captureEvent(
+            {
+              // @ts-ignore __INTERNAL_MINIDUMP is not assignable to event
+              __INTERNAL_MINIDUMP: true,
+            },
+            {
+              data: {
+                __sentry__: true,
+              },
```
HazAT, line 158: if an event has this in the hint, we will not invoke beforeSend.

jan-auer: Do we even need the `__INTERNAL_MINIDUMP` then?

HazAT: Yes we do, this `__sentry__` is used internally if our SDK captures something itself.
The `__INTERNAL_MINIDUMP`, while being a workaround is something else.

vyunikov:
Posted my comment in https://github.com/getsentry/sentry-electron/issues/249
Does it ever gets called in case of a crash in main process?

Issue #200, opened 2019-12-06
Minidumps on Linux are Multipart Bodies


jan-auer:
## Versions + Platform

- [x] SDK version - all
- [x] Electron version - all
- [x] Platform - `Linux`

## Description

The Electron crash reporter stores the entire multipart request body into the `.dmp` file. The Electron SDK then uploads this file as Minidump and wraps it in another Multipart request. While Sentry can deal with it, it is very strange behavior and creates unnecessarily complex requests. Instead, the SDK could send the file directly and just append the `sentry` payload to the body. 

To do this, the SDK would need to:
- Open the file and parse the first line to obtain the multipart boundary
- Prepend the file with another boundary, and the `sentry` field (see below for an example).
- Send the result directly as request body. This requires a manual header `Content-Type: multipart/form-data;boundary="boundary"`

The result should look like this:

```
--boundary 
Content-Disposition: form-data; name="sentry" 
Content-Type: application/json

{"event_id":"feedbeef","user":{}}
--boundary 
...
```

PR #278, opened 2020-12-08
Handle multipart/form-data submissions on native crashes in Linux

implausible:
Fixes https://github.com/getsentry/sentry-electron/issues/200. I think what was not mentioned in the issue is that currently Sentry _does not handle_ these submissions at all, and so native crash reports on Linux are currently being labeled as corrupted submissions in Sentry. We need this solved ASAP so that we can adopt the product.

This adds the ability to detect whether a crash report is a `multipart/form-data` submission body and to switch minidump upload behavior to accommodate that on Linux. I have the `multipart/form-data` being sent to the `/api/{projectId}/minidump` route instead of the `/api/{projectId}/envelope` route, as when I tried uploading the minidump with the sentry event attached to it, sentry's servers returned a 400 status code, causing the system to throw. If it was intended to allow the envelope route to accept the `multipart/form-data` submission, it might need to be fixed server side to return a 200 OK in response.

I have adjusted the `multipart/form-data` submission to include the event payload under the name `sentry` as suggested in the issue. From testing against sentry, I can confirm that this produces the desired behavior where the full sentry event and minidump attachment are properly received on sentry.

Due to the changes introduced here, I went ahead and adjusted the test server to accept minidump `multipart/form-data` submissions. I also reenabled all tests as well as addressed the blocking issues from running tests on 10 and up. I was able to reproduce the issues that caused the team to disable specs on `9.0.5`. I traced the failure to this issue https://github.com/electron/electron/pull/24788. This was addressed in `9.2.0`. I would advise that for anyone affected by that issue, the team suggests to the affected user to move their Electron target.

Please let me know if there is anything I can do to speed up the process of getting this merged / released. My team is looking to adopt Sentry for crash reporting and need this solved.


implausible:
I've disabled testing on Electron < 4 due to some stability issues that exist in the windows e2e tests. While it's not really my decision what your minimum support is, Electron has a much smaller support target than it maintained in the past, with versions < 9 no longer being supported by the team.

timfish:
I was doing some Electron v11 testing locally today and saw the same test `onFatalError` test failure on macOS too.

implausible:
> I was doing some Electron v11 testing locally today and saw the same test onFatalError test failure on macOS too.

I am not sure that this is a fault of this branch though. From an improvement perspective, I believe this branch covers a condition that the sentry electron sdk currently does not support without introducing regressions. Are there any changes you would like me to make? I'd love to see this merged and released, so that I can stop relying on a fork for sentry support in Electron 10.

test/e2e/index.test.ts
```diff
@@ -250,7 +240,12 @@ describe('E2E Tests', () => {
         expect(event.data.fingerprint).to.include('abcd');
       });
 
-      it('Loaded via preload script with nodeIntegration disabled', async () => {
+      it('Loaded via preload script with nodeIntegration disabled', async function() {
+        if (majorVersion < 8) {
```
timfish, line 244: I think we can probably leave this test enabled for all Electron versions as they are passing in the other PRs.

timfish:
The preload tests are failing due to [this erroneous line](https://github.com/getsentry/sentry-electron/pull/286/files#diff-fb98f75280477dd7fc014c9d5f4eff3a59083d7ae9c64e3989aeb47a1b4b05bcL7) which I think was left on master by mistake. I've removed it in my recent un-merged PRs and it fixes the failures.

timfish:
@implausible if you can rebase this to master I can get it merged!

implausible:
I will get to this today

timfish:
I managed to do the merge via GitHub web UI. Just looking into tests...

implausible:
> I managed to do the merge via GitHub web UI. Just looking into tests...

Oh, awesome!

timfish:
@implausible it looks like there are many failing tests for minidumps. 

Some debugging shows that the envelope body is not parsed correctly by the test server and this looks to be caused by the changes to `net.ts` passing the content type.

PR #280, opened 2020-12-10
feat: Cater for contextIsolation

### LINKED ISSUE #279: Renderer Backend will not work without access to ipcRenderer
Electron's security [documentation](https://www.electronjs.org/docs/tutorial/security) suggest setting:
- `nodeIntegration` to false
- `contextIsolation` to true
- using `contextBridge` (https://www.electronjs.org/docs/api/context-bridge, https://stackoverflow.com/a/57656281/6319004)

The sentry client for the renderer directly `requires` `ipcRenderer` (https://github.com/getsentry/sentry-electron/blob/master/src/renderer/backend.ts#L5) as a way to send events to the main process, but with the suggested security settings, this would not be possible (`ipcRenderer` wouldn't be available in the renderer process).

Are there any plans to allow us to use sentry in the renderer process without exposing `ipcRenderer` directly?

If not, would it be possible for us to submit a PR for this?

timfish:
Fixes #279 

- Attaches IPC methods to `window` so non-isolated context uses the same code paths as isolated
- Uses `contextBridge` when required/available
- Adds test app with `contextIsolation: true`
- Tests building test app via webpack
- Tests events are working with `contextIsolation: true`


src/main/backend.ts
```diff
@@ -188,7 +188,7 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
     // error.
     const dsnString = this._options.dsn;
     if (!dsnString) {
-      throw new SentryError('Invariant exception: install() must not be called when disabled');
```
timfish, line 191: I modified this error message because I hit it in testing when not passing a DSN and it didn't point me in the right direction.

timfish:
There are many renderer tests failing for Electron v10 and 11 on Linux and Windows. Need to investigate but it seems unlikely that they are related to changes in this PR.

EvanPurkhiser:
@HazAT any reason we haven't taken a look here?

bermanboris:
Any news on this PR?

HazAT (approved):

PR #286, opened 2020-12-17
feat: Compress uploads

### LINKED ISSUE #285: Minidumps are not compressed before upload

## Versions + Platform

- [x] SDK version - `@sentry/electron@v2.0.4`
- [x] Electron version - `electron@v8.2.4`
- [x] Platform - `Windows 10`

## Description

Documentation mentions 20MB limit for a compressed minidump request: https://docs.sentry.io/platforms/native/guides/minidumps/#size-limits

But it seems my minidump request is not beeing compressed. Sentry successfully uploads ~19MB minidumps on app restart (after crash), but fails to upload 24 MB minidump **(Sentry Logger [Warn]: Failed to upload minidump SentryError: HTTP Error (413))**. Why are you not compressing minidumps? Documentation says you support compressed minidump uploads. After compression, my 24MB minidump becomes a mere 240 KB zip file.

timfish:
Fixes #285

### DIFF
--- a/src/main/transports/net.ts
+++ b/src/main/transports/net.ts
@@ -3,6 +3,7 @@ import { Event, Response, SentryRequest, Status, TransportOptions } from '@sentr
 import { parseRetryAfterHeader, PromiseBuffer, SentryError, timestampWithMs } from '@sentry/utils';
 import { net } from 'electron';
 import * as url from 'url';
+import { gzip } from 'zlib';
 
 import { isAppReady } from '../backend';
 
@@ -84,6 +85,7 @@ export class NetTransport extends Transports.BaseTransport {
         const options = this._getRequestOptions(new url.URL(request.url));
         options.headers = {
           ...options.headers,
+          'Content-Encoding': 'gzip',
           'Content-Type': 'application/x-sentry-envelope',
         };
         const req = net.request(options as Electron.ClientRequestConstructorOptions);
@@ -117,8 +119,11 @@ export class NetTransport extends Transports.BaseTransport {
             // Drain
           });
         });
-        req.write(request.body);
-        req.end();
+
+        gzip(request.body, (_, compressedBody) => {
+          req.write(compressedBody);
+          req.end();
+        });
       }),
     );
   }
--- a/src/main/transports/net.ts
+++ b/src/main/transports/net.ts
@@ -8,6 +8,7 @@ import { createGzip } from 'zlib';
 
 import { isAppReady } from '../backend';
 
+// Estimated maximum size for reasonable standalone event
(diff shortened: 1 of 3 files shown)


src/main/transports/net.ts
```diff
@@ -117,8 +119,11 @@ export class NetTransport extends Transports.BaseTransport {
             // Drain
           });
         });
-        req.write(request.body);
-        req.end();
+
+        gzip(request.body, (_, compressedBody) => {
+          req.write(compressedBody);
```
jan-auer, line 124: This is fine. If you like to reduce overheads here, you could also create a gzip stream and then have it flush into the request, like described [here](https://nodejs.org/api/zlib.html#zlib_flushing).

jan-auer (approved):
Thank you, @timfish! LGTM. 

As a suggestion, you could only trigger compression if the body exceeds a certain size, since gzip also comes with a CPU penalty.

timfish:
> if the body exceeds a certain size

What would be a good threshold?

jan-auer:
I did some quick research, and without explicit benchmarking the threshold is hard to determine. Most resources I found talk about a limit lower than 1k, which would be almost every outgoing request.

There is also https://medium.com/@joehonton/the-gzip-penalty-d31bd697f1a2, which seems to tip over at 32k. This gets me thinking: We probably don't need compression for regular event submission, and events are usually under 32k, so let's go with that?

HazAT:
So `if(request.body.length > 32000) zip else not zip`?

timfish:
Now with streams and gzip threshold!

src/main/transports/net.ts
```diff
@@ -2,17 +2,34 @@ import { Transports } from '@sentry/node';
 import { Event, Response, SentryRequest, Status, TransportOptions } from '@sentry/types';
 import { parseRetryAfterHeader, PromiseBuffer, SentryError, timestampWithMs } from '@sentry/utils';
 import { net } from 'electron';
+import { Readable, Writable } from 'stream';
 import * as url from 'url';
+import { createGzip } from 'zlib';
 
 import { isAppReady } from '../backend';
 
+const GZIP_THRESHOLD = 1024 * 32;
```
jan-auer, line 11: _nit_: I'd mention in a comment that this is an estimated maximum size for reasonable standalone events.

jan-auer (approved):
Looks great, thanks for going the extra mile.

GintV:
Any particular reason why this one is not merged yet?

GintV:
Thanks for implementing this @timfish! @jan-auer, @HazAT any chance release could be made with this change?

PR #297, opened 2021-02-08
meta: bump sentry dependencies to lates 6.1.0 version

kevinlinv:
## Descriptions

This PR upgrades to the latest Sentry version.

I tried running the e2e test locally and it failed on electron 11 for `onFatalError can be overridden` both before and after this bump. So I think it is unrelated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,17 +37,17 @@
     "e2e": "cross-env TS_NODE_PROJECT=tsconfig.json xvfb-maybe mocha --require ts-node/register/transpile-only --timeout 30000 --retries 5 ./test/e2e/*.ts"
   },
   "dependencies": {
-    "@sentry/browser": "5.27.6",
-    "@sentry/core": "5.27.6",
-    "@sentry/minimal": "5.27.6",
-    "@sentry/node": "5.27.6",
-    "@sentry/types": "5.27.6",
-    "@sentry/utils": "5.27.6",
+    "@sentry/browser": "6.1.0",
+    "@sentry/core": "6.1.0",
+    "@sentry/minimal": "6.1.0",
+    "@sentry/node": "6.1.0",
+    "@sentry/types": "6.1.0",
+    "@sentry/utils": "6.1.0",
     "tslib": "^1.9.3"
   },
   "devDependencies": {
-    "@sentry-internal/eslint-config-sdk": "5.27.6",
-    "@sentry-internal/typescript": "5.27.6",
+    "@sentry-internal/eslint-config-sdk": "6.1.0",
+    "@sentry-internal/typescript": "6.1.0",
     "@types/body-parser": "^1.17.0",
     "@types/chai": "^4.2.10",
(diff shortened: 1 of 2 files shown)


kevinlinv:
Heya @HazAT we would love to get this version bump to leverage new features on the latest release. Would love it if you could have a look and perhaps merge this PR 🙂

timfish (approved):
The tests can be flakey and need investigating but this looks good.

kevinlinv:
Hey, thanks for the review 🎉 
Would be lovely if someone with write permission could help merge this 🙂

HazAT:
@kevinlinv Thanks for the PR!
We need to be slightly more careful here since starting from 6.x we enabled sending sessions by default.
We didn't spend the time yet to figure out what this means for electron and how this will work in our product.
There is no real reason to update the dependencies besides this.

kevinlinv:
Hey, thanks for the quick update 👍  

For context, the reason I am filing this PR here is because we're currently operating on a mono repo setup, where we have `@sentry/browser|integration|types|electron` installed as dependencies, we'd like to update to latest `v6` to leverage some new feature on `@sentry/browser|integration|types`. But are running into an incompatible types when using an integration from `@sentry/integrations@6.x.x` with `@sentry/electron@latest` which still expects `@sentry/integrations@5.27.6`.

If updating the dependencies in here isn't straight forward and might not make sense, we may have to find another way to deal with the types incompatibility on our end 👌

lobsterkatie:
> we'd like to update to latest v6 to leverage some new feature on @sentry/browser|integration|types

@kevinlinv - what new things from `6.x` are you hoping to use?

kevinlinv:
Hi @lobsterkatie thanks for the follow up.

We're keen on giving release health feature https://docs.sentry.io/product/releases/health/ 
We're still blocked on this upgrade though, unfortunately.

lobsterkatie:
Release health is available starting in 5.27.0. The only difference between that and 6.x is that session tracking is turned on by default in the latter. See https://github.com/getsentry/sentry-javascript/pull/2973.

kevinlinv:
Hey @lobsterkatie thanks, thats news to me, there may have been a misunderstanding on my part.
Will give it a go to enable the feature through `init` config in `5.27.6` and shall close this if we can get it to work!

aguynamedben:
Looking for this to go to a 6.x version so I can use the tracesSampler option.

Related: https://github.com/getsentry/sentry-javascript/issues/3324#issuecomment-806192264

kevinlinv:
Not actively looking at this PR anymore, I'll close this so that others could re-propose the update if needed.

lobsterkatie:
> Looking for this to go to a 6.x version so I can use the tracesSampler option.

@aguynamedben - you can - it was [added in 5.24](https://github.com/getsentry/sentry-javascript/commit/7c142839036e77161a6f707196762eb718b723cc).

PR #302, opened 2021-03-01
Update backend.ts

amcode21:
Use app.getPath('crashDumps') to not break in electron v12


timfish:
Thanks for the PR.

We want to continue supporting older versions of Electron so I'l close this in favour of #303

PR #304, opened 2021-03-03
feat: Improve tests

timfish:
In its default configuration, `@sentry/electron` uses its own nodejs based uploader. This allows us to attach detailed context along with captured minidump files. This is not possible (or at least very difficult) with the native uploader built into Electron because: 
- It only supports 127 byte values for extra parameters on Windows
- Extra parameters must be added from both the main and renderer if you want it to be transmitted with crashes from both

The main downside with the nodejs uploader is that if the main process encounters a native crash before the uploader can run, no dump will be sent. 

The proposed workaround for this is:
- Start `crashReporter` with `uploadToServer: true`
- Once the main process is up and somewhat "stable", enable the nodejs uploader and call `crashReporter.setUploadToServer(false)` to disable the native uploader

This should ensure that crashes are sent regardless of whether the nodejs uploader gets to run.

**Downsides**
Native crashes in the main process will always be sent via the native Electron uploader without full context. 

- [x]  Reinstate `minidump` test endpoint and tests
- [ ] ~~Decide how we might determine when to switch to nodejs uploader~~
- [ ] ~~Code the switchover~~
- [ ] ~~Add tests!~~


timfish:
I'm now not convinced of the value gained by switching uploader at run-time so I'm going to constrain the scope of this PR to simply improving the tests.

timfish:
So to summarise this PR:
- Changes test server to use Koa. I found it easier to get compressed multi-part uploads working.
- Add minidump test endpoint to handle Electron uploader
- Added tests for Electron uploader
- Removed unnessary test
- Merged a couple of tests

HazAT (approved):

PR #311, opened 2021-03-09
Fix: improve native uploader context

timfish:
The Electron native crash uploader only allows short strings to be attached for context and these are limited to 127 bytes on Windows.

This PR allows you to pass a function as configuration option and this function is used to transform/extract parts of the scope to be sent with the uploads.

#309 was required to get this working. It fixed a bug where scope from the main process was being overwritten with empty scope from renderers.

For example, to enable the native uploader, disable the sentry uploader and select only the `user` object from the scope, you would initialse the SDK like this:

```typescript
init({
  // ...
  useCrashpadMinidumpUploader: scope => {
    const user = scope.getUser();
    return { user };
  },
  useSentryMinidumpUploader: false,
  // ...
});
```

This small amount of extracted scope is passed to the native crash reporter in both the main and renderer process so it gets passed with either type of crash.

Note that the scope extraction function is only called from the main process so you have full access to nodejs and can extract scope depending on `process.platform` etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@sentry/node": "5.27.6",
     "@sentry/types": "5.27.6",
     "@sentry/utils": "5.27.6",
-    "tslib": "^1.9.3"
+    "tslib": "^2.1.0"
   },
   "devDependencies": {
     "@sentry-internal/eslint-config-sdk": "5.27.6",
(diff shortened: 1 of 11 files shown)


timfish:
Tests are failing on macOS with Electron <=8 for renderer scope with Electron uploader. I'm guessing this is because it was only at v9 where extra parameters were correctly applied in the renderer process.

Similar failures were observed on Windows too so they are already disabled. This means the full extent of this feature is only available on Electron >= v9.

timfish:
Clsoing in favour of #315

PR #315, opened 2021-03-10
feat: Context and breadcrumbs sent via Crashpad for native crashes in main process when `useCrashpadMinidumpUploader` enabled

timfish:
When using the native Electron uploader, extra context can be passed by calling `crashReporter.addExtraParameter(key, value)` and Sentry automatically reads the event from the `sentry` form field as JSON.

Windows and macOS truncate extra parameters over 20320 bytes and 100 breadcrumbs can easily cause the JSON to be over 20k. For this reason we make use of an undocumented Sentry backend feature where `sentry__x` parameters (where x is ascending numbers) are concatenated before being parsed as JSON.

For Linux extra parameters are limited to 127 byes and breakpad has built in support for auto chunking but it's limited to 160*127 byte parameters (20320 bytes total). ~~For this reason we manually chunk on Linux too because then it's possible to exceed the 20k limit.~~ We're not supporting Linux for the moment.

Summary of changes:
- Add parameter concatenation to test server
- Moved `SdkInfo` from `_prepareEvent` into default context so its available from backend
- Put `IPC` strings behind a single property
- ~~Added `EXTRA_PARAMS` IPC channel to send chunked params to renderer~~
- On main process scope change:
  - Create dummy event with scope + context
  - Split event JSON so it fits into native uploader params
  - Add params to native uploader in main process
  - ~~Send params to renderers and add them to native uploader~~
- Update tests to check context is passed with Electron uploader (only Electron v9-12)


timfish:
Linux auto parameter chunking is limited to 160 * 127 byte paramters (20320 bytes). This fails with 100 breadcrumbs.

src/main/backend.ts
```diff
@@ -150,30 +142,95 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
...
+   * Returns chunks with keys and values:
+   * {
+   *    sentry__1: '{ json...',
+   *    sentry__2: 'more json...',
+   *    sentry__x: 'end json }',
+   * }
+   */
+  private _getNativeUploaderExtraParams(event: Event): { [key: string]: string } {
+    // We chunk on Linux too because the built in chunking is limited to 160 * 127 byte parameters
+    const maxBytes = process.platform === 'linux' ? 120 : 20300;
```
jan-auer, line 204: This will change once Electron uses Crashpad on Linux. Let's keep this in mind so we can introduce an Electron version switch in the future.

src/main/backend.ts
```diff
@@ -150,30 +142,95 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
...
-              // @ts-ignore __INTERNAL_MINIDUMP is not assignable to event
-              __INTERNAL_MINIDUMP: true,
-            },
-            {
-              data: {
-                __sentry__: true,
-              },
-            },
+          forget(
+            this._getNativeUploaderEvent(scope).then(event => {
```
jan-auer, line 154: This starts an asynchronous operation. Can we prevent race conditions from consecutive updates?

src/main/backend.ts
```diff
@@ -150,30 +142,95 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
...
+              // Update the extra parameters in the main process
+              const mainParams = this._getNativeUploaderExtraParams(event);
+              for (const key of Object.keys(mainParams)) {
+                crashReporter.addExtraParameter(key, mainParams[key]);
+              }
+
+              // Modify the event to point to renderer crash
+              event.extra = { ...event.extra, crashed_process: 'renderer' };
```
jan-auer, line 162: _nit_: Might be faster to just overwrite the one field, assuming `extra` is always set:

```suggestion
              event.extra.crashed_process = 'renderer';
```

src/main/backend.ts
```diff
@@ -150,30 +142,95 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
...
+              // Update the extra parameters in every renderer process
+              const rendererParams = this._getNativeUploaderExtraParams(event);
+              for (const window of BrowserWindow.getAllWindows()) {
+                if (!window.webContents.isDestroyed()) {
+                  window.webContents.send(IPC.EXTRA_PARAMS, rendererParams);
```
jan-auer, line 167: Do we absolutely need to do this broadcast? It looks like `initialScope` covers the most important case to have shared scope between _all_ processes.

src/main/backend.ts
```diff
@@ -237,11 +294,13 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
       contents: Electron.WebContents,
       details?: Electron.RenderProcessGoneDetails,
     ): Promise<void> => {
-      try {
-        await this._sendNativeCrashes(this._getNewEventWithElectronContext(contents, details));
-      } catch (e) {
-        // eslint-disable-next-line no-console
-        console.error(e);
+      if (this._options.useSentryMinidumpUploader !== false) {
```
jan-auer, line 297: This is purely for my education: How does this flag relate to `useCrashpadMinidumpUploader`?

jan-auer (commented):
> it's limited to 160*127 byte parameters (20320 bytes total). For this reason we manually chunk on Linux too because then it's possible to exceed the 20k limit.

How does manual chunking avoid the 20k limit? You're still limited to 160 keys in total.

timfish:
I need to check this again on my Linux machine when I get back this evening.

Auto chunking appeared to just stop at 160 keys rather than there be a 160 key limit.

src/main/backend.ts
timfish: I'm not entirely sure since I didn't add them!

My best guess is that it's so you can have minidump capture enabled but disable uploading with both set to false.

src/main/backend.ts
timfish: If you want breadcrumbs and full context with native renderer crashes via the Electron uploader, this is the only way to do that.

My understanding was that `initialScope`/`globalExtra` was to add minimal context for GPU crashes which are not covered by the above.

src/main/backend.ts
timfish: Good point 🤔

src/main/backend.ts
timfish: If `initialScope` is enough context for ALL native crash reports via the Electron uploader, we can pull this PR entirely.

@HazAT added usage of `addExtraParameter` a year ago [here](https://github.com/getsentry/sentry-electron/pull/244/files#diff-39928ad64d67bbb28adaeb7b5e0b6580af610e5b19d5a3cf730d59c39c7bffa1R118) so I had assumed full context was a requirement. There are a couple of issues with the previous implementation:
- `addExtraParameter` called in `sendEvent` so context would only get added after an event
- `addExtraParameter` needs calling in the renderer if you want context there

timfish:
@jan-auer Here is why Electron breakpad on Linux is limited to 160*127 bytes for auto chunking:
https://github.com/electron/electron/blob/a1a649c0d610767667a5fd7039772cbab237f90c/shell/common/crash_keys.cc#L63-L78

Looks like they start from the end chunk and work their way backwards.

There is no hard limit that I've found for the number of keys which is why manually chunking works beyond the 20320 limit.

Let me know if you want to support context other than `initialScope` in native crashes via the Electron uploader. If not I'll close this and have a go rebasing #334 off master!

src/main/backend.ts
timfish: `extra` does appear to always be set but it's optional in the type definitions which means TypeScript throws an error with  this:
```ts
event.extra.crashed_process = 'renderer';
```

timfish: oh, this code is going anyway

src/main/backend.ts
```diff
@@ -163,17 +168,102 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
...
+          const mainParams = this._getNativeUploaderExtraParams(event);
+          for (const key of Object.keys(mainParams)) {
+            crashReporter.addExtraParameter(key, mainParams[key]);
+          }
+        } finally {
+          this._updatingExtraParams = false;
+        }
+      })(),
+    );
+  }
```
timfish, line 211: @jan-auer does this look better to avoid race conditions?

jan-auer: I'm afraid this could now result in lost updates.

I took a closer look now, and the culprit is that event defaults can be initialized lazily. Changing that is not practical in this PR, so I suggest a small change: Ensure that always the _last_ update comes through instead of the first one. Something like:

```javascript
const currentEpoch = ++this.updateEpoch;

forget(async () => {
  await blablabla();
  if (currentEpoch !== this.updateEpoch) return;
  addExtraParameter(...);
});
```

jan-auer (commented):
Looks good to me now, thanks for bearing with me. 

A final suggestion on how to resolve the race condition, and then we can merge this 👍

src/main/backend.ts
timfish: Yeah, this is a much better idea!

jan-auer (approved):

PR #319, opened 2021-04-14
Check for require with typeof

k15a:
If require is undefined the following code checking for the existence of require will throw a reference error so the nicer custom error message is never displayed to the user.

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -118,7 +118,7 @@ declare interface CrossApp extends App {
 
 /** Get the name of an electron app for <v5 and v7< */
 export function getNameFallback(): string {
-  if (!require) {
+  if (typeof require === 'undefined') {
     throw new SentryError(
       'Could not require("electron") to get appName. Please ensure you pass `appName` to Sentry options',
     );


timfish (approved):

src/common.ts
```diff
@@ -118,14 +118,16 @@ declare interface CrossApp extends App {
 
 /** Get the name of an electron app for <v5 and v7< */
 export function getNameFallback(): string {
-  if (!require) {
+  let electron;
+
```
timfish, line 122: @k15a I've got another PR in the works and this could make it into a release soon. 

Could you add some comments here to explain about esbuild and why this is required? There's always commit history but this looks like the sort of code that I would immediately clean up without a second thought!

timfish (changes requested):

src/common.ts
k15a: Hi, ESBuild has removed the warning but Evan correctly pointed out that this code was not doing anything before. If require is not defined the check will already throw a ReferenceError and your helpful error message would never reach the user. I will add a comment saying that `require` can be not defined and throwing a reference error.

Screen Shot 2021-04-22 at 09 41 26

timfish:
Thanks @k15a

PR #334, opened 2021-05-26
feat: Add `initialScope` to `globalExtra`

timfish:
This PR also contains required changes from #315. If that's merged into master first, it'll be easier to see these minor changes.

- Converts `initialScope` to JSON and adds as `globalExtra` with native crashReporter 
- Modifies test server and tests to check `sentry___initialScope` is passed through 

> Note: If the `globalExtra` JSON exceeds ~30kB, the crashReporter appears to silently fail to send the minidump!


timfish:
The Windows tests pass locally but sometime hang in CI 🤷‍♂️

src/main/backend.ts
```diff
@@ -72,7 +72,7 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
 
     if (this._isNativeEnabled()) {
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      forget(this._installNativeHandler());
+      forget(this._installNativeHandler(options.initialScope));
```
jan-auer, line 75: The Core SDK already handles this option and writes it into the current hub's scope. It would be great if you just read from that scope, rather than handling the option another time. 

However, this means you'd have to pay attention to the sequence of initialization, because this has to run _after_ the Core SDK has initialized the scope.

timfish:
Closed in favour of #340 which is based off master.

Issue #205, opened 2019-12-25
Huge amount of breadcrumbs


ghost:
Current `"@sentry/electron": "1.1.0"` is sending unlimited amount of breadcrumbs:
![image](https://user-images.githubusercontent.com/56207736/71427328-8e3e8c80-26c8-11ea-8470-a5746a9e1c16.png)

I'm using Electron 6.1.5, and I'm using Sentry from the main and two renderer processes.

kr-deps:
Rolled back to `"@sentry/electron": "0.17.3"`, sends only 21 breadcrumbs now as it used to.

mateo-io:
I can confirm.

I develop a big electron app and the latest Sentry was causing big performance problems. Downgrading to `0.17.3` and they are no longer noticeable.

williamweckl:
Anyone knows if this is still happening?

williamweckl:
I think this is releated to #277. I've tested here and don't think this is still happening.

PR #336, opened 2021-05-27
Fix: Few minor improvements

### LINKED ISSUE #181: Default breadcrumbs are getting a bit noisy
As of Electron v4, the default breadcrumbs from Electron events are getting a bit noisy. We probably need to exclude `app.remote-get-builtin`, `app.remote-get-global` and `app.remote-get-current-window`:

image

### LINKED ISSUE #222: Could not connect to Sentry main process. Did you call init?
`
Could not connect to Sentry main process. Did you call init?
`

But I did init it !!

### LINKED ISSUE #277: Memory leak: breadcrumbs accumulating
The buggy line:
https://github.com/getsentry/sentry-electron/blob/942b1f21da743a975701bed46b8d1603c9d7193e/src/main/backend.ts#L309

The call needs the second argument `maxBreadcrumbs`. Currently breadcrumbs can accumulate endlessly through this particular call.

I'll submit a PR if I find out how to get `maxBreadcrumbs` from there.

timfish:
- Fixes #222 by making the error text more clear
- Fixes #277 and #205 by ensuring `maxBreadcrumbs` is passed (This doesn't currently appear be an issue but I see no reason good reason not to pass it)
- Fixes #181 by not capturing `remote-` `app` events
- Calls `taskkill /f /im electron.exe` after each test on Windows to try and make CI get stuck less on hanging Electron processes
- Adds Electron 13 to e2e tests


jan-auer (commented):
Changes look good to me, with one exception: @kamilogorek can you verify the change on `scope.addBreadcrumb`? I think this shouldn't be necessary, since this is a public API that we also don't expect users to call with maxBreadcrumbs. The scope should know this already.

kamilogorek:
Unfortunately, it is necessary right now when called directly on the scope instance. Hub knows about the client configuration, but not the scope. There's an open issue for this as well - https://github.com/getsentry/sentry-javascript/issues/2963

src/main/backend.ts
```diff
@@ -326,9 +326,9 @@ export class MainBackend extends BaseBackend<ElectronOptions> implements CommonB
           scope.setExtras(sentScope._extra);
         }
 
-        // Since we do not have updates for individual breadcrumbs anymore and only for the whole scope
+        // Since we do not have updates for individual breadcrumbs any more and only for the whole scope
         // we just add the last added breadcrumb on scope updates
-        scope.addBreadcrumb(sentScope._breadcrumbs.pop());
+        scope.addBreadcrumb(sentScope._breadcrumbs.pop(), this._options.maxBreadcrumbs);
```
timfish, line 331: @kamilogorek is it required to pass `maxBreadcrumbs` down like this? 

There are a few open issues (#277, #205) about excessive breadcrumbs and although I haven't seen this myself in a long time, it seems like this might be required?

timfish: 🤦‍♂️ Just seen you've already replied...

kamilogorek: Tbh this https://github.com/getsentry/sentry-javascript/blob/master/packages/hub/src/hub.ts#L45-L49 should be moved to `scope.ts`, and logic could be slightly modified in both places (`hub/scope`), which would at least prevent those overacumulated leaks.

PR #340, opened 2021-06-14
feat: Add `initialScope` to `globalExtra`

timfish:
- Updates the e2e test server to support latest Sentry Relay features
- Adds `initialScope` as JSON to Electron `globalExtra` form parameters as `sentry___initialScope` 
- Adds a GPU crash test for Electron >= 13 with Electron uploader


src/main/backend.ts
```diff
@@ -41,6 +41,26 @@ export async function isAppReady(): Promise<boolean> {
   );
 }
 
+/** Gets a Scope object with user, tags and extra */
+function getScope(): Partial<ScopeContext> | undefined {
```
timfish, line 45: Is there a better way than this to get an object with `user`, `tags` and `extra` from the scope?

kamilogorek: Not really, although I'd still rewrite it as:

```js
function getScope(): Partial<ScopeContext> {
  const scope = getCurrentHub().getScope() as any;

  if (!scope) {
    return {};
  }

  /* eslint-disable @typescript-eslint/no-unsafe-member-access */
  return {
    ...(scope._user && { user: scope._user }),
    ...(scope._tags && { tags: scope._tags }),
    ...(scope._extra && { extra: scope._extra }),
  };
}
```

It should be totally safe to always set initial scope to `{}` on line 220 then as well.

timfish:
Looks like I need to do some debugging on Linux

timfish:
@jan-auer I seem to have hit a rather strange bug in Breakpad where the `globalExtra` values get truncated at the first comma character which is quite a pain for JSON!

Do you want to just limit this feature to Crashpad platforms for now and we'll come back to this later?

jan-auer:
Agreed, let's do just crashpad and then enable Linux when it has switched over.

nornagon:
👋  thanks for this change! Super helpful to have the scope available on renderer crashes.

One more thing though—it seems like fields like `release` aren't automatically propagated here, so e.g. when searching for `release:xxxx`, all renderer crashes will be excluded. Would it be possible to include that data in `globalExtra` too?

timfish:
Yes, good point. I'll get that added!

PR #357, opened 2021-07-26
feat: Support Crashpad on Linux

timfish:
Electron now supports [Crashpad on Linux](https://github.com/electron/electron/issues/27859).  

Enables Crashpad features on Linux when Electron >= v16 or >= v15 with switch enabled.


timfish:
These changes have been rolled into #361

Issue #355, opened 2021-07-23
Renderer fails on 'fs' requirement


umerkk:
## Version
    "@sentry/electron": "^2.5.1",

I am integration Sentry to an Electron app using this SDK, so far the Main Process seems to be working fine & sending the event to sentry, but the renderer doesn't work.

I always throw an error 

`Module not found: Error: Can't resolve 'fs' in '/Users/muhammad.umer/Repo/hub.uw-hub/node_modules/electron'
 @ ./node_modules/electron/index.js 1:9-22
 @ ./node_modules/@sentry/electron/esm/renderer/backend.js
 @ ./node_modules/@sentry/electron/esm/renderer/index.js
 @ ./src/renderer/features/downloadDrawer/DownloadDrawer.tsx
 @ ./src/renderer/layout/AppLayout.tsx
 @ ./src/renderer/App.tsx
 @ ./src/renderer/index.tsx
 @ multi ./src/renderer/index.tsx`

I am initializing in the Preload script of my renderer(Browser Window).

##### Preload.js
const Sentry = require('@sentry/electron/dist/renderer');
Sentry.init({ dsn: 'https://000@000.ingest.sentry.io/000' });

I am explicitly referencing the renderer dist from the SDK in my renderer process, but don't seem to understand why it is referring to nodeFS.

BrowserWindow options
`        webPreferences: {
          textAreasAreResizable: false,
          preload: path.join(__dirname, 'preloads', 'mainWindowPreload.js'),
          enableRemoteModule: true,
        },`

timfish:
There is no usage of `fs` in the SDK renderer code. 

The supplied stacktrace seems to suggest that `fs` is being called from `node_modules/electron/index.js`

Do you have a repository that reproduces this? None of the existing tests or examples exhibit this behaviour.

umerkk:
It's Strange because I do not get these errors if i remove Sentry SDK usage from my codebase (Production code).
There must be some dependency using 'fs' using the probably remote module of the electron.

I was able to at least compile the code by changing the webpack `target: from from `web` to  'electron-renderer`.
Although it still requires `node_modules` to be part of the renderer dist otherwise it doesn't work and I am excluding `node_modules`.

Now after making those above changes, I still get the `fs` error in console.log of the browser but the events are appearing in Sentry.

My Renderer is in 'React + Typescript' and Main in JS + TS.

This is my webpack config.
`const config: WebpackConfiguration = {
[...]

Issue #351, opened 2021-07-14
init() crashes in preload script


Achder:
- [x] Review the documentation: https://docs.sentry.io/platforms/javascript/electron/
- [x] Search for existing issues: https://github.com/getsentry/sentry-electron/issues
- [x] Use the latest release: https://github.com/getsentry/sentry-electron/releases
- [ ] Provide a link to the affected event from your Sentry account

## Versions + Platform

- [x] SDK version - `@sentry/electron@v2.5.1`
- [x] Electron version - `electron@v13.1.6`
- [x] Platform - `Windows`

## Description

I'm currently updating my app to use electron 13 and trying to apply proper context isolation and disabling node integration.

```
webPreferences: {
	webSecurity: isDevelopment ? false : true,
	nodeIntegration: false,
	contextIsolation: true,
	worldSafeExecuteJavaScript: true,
	spellcheck: false,
	preload: path.resolve(path.join(app.getAppPath(), 'preload.js')),
	defaultFontSize: 14,
}
```

I know from other issues postet here, that I need to call init() in the main process, my preload script and the isolated renderer.
But my preload script crashes with the following error:

```
VM75 renderer_init.js:93 TypeError: Cannot read property 'getAppPath' of undefined
    at eval (webpack://***/./node_modules/@sentry/electron/esm/main/normalize.js?:10:58)
    at Module../node_modules/@sentry/electron/esm/main/normalize.js (C:\Development\***\dist\main\preload.js:355:1)
    at __webpack_require__ (C:\Development\***\dist\main\preload.js:5085:33)
    at fn (C:\Development\***\dist\main\preload.js:5244:21)
    at eval (webpack://***/./node_modules/@sentry/electron/esm/main/backend.js?:25:69)
    at Module../node_modules/@sentry/electron/esm/main/backend.js (C:\Development\***\dist\main\preload.js:278:1)
    at __webpack_require__ (C:\Development\***\dist\main\preload.js:5085:33)
    at fn (C:\Development\***\dist\main\preload.js:5244:21)
    at eval (webpack://***/./node_modules/@sentry/electron/esm/main/client.js?:9:66)
    at Module../node_modules/@sentry/electron/esm/main/client.js (C:\Development\***\dist\main\preload.js:289:1)
```

getAppPath() is only available in the main process.

I'm using webpack as a bundler and selected `electron-preload` as my target for the preload script:
```
merge(common, {
	mode: 'development',
	target: 'electron-preload',
[...]

PR #359, opened 2021-07-29
Automatically inject preload scripts

timfish:
With this PR, we automatically inject the correct preload scripts via the Electron `session.setPreloads()` API. 

This has a number of advantages:
- Lots of code removed and simplified
- Users no longer have to supply `appName` in any process
- Users no longer have to consider preload scripts unless they're bundling the main process
- When bundling, no more issues with incorrect target for preload (#351)
- Renderer code no longer needs to reference Electron so no need to add `electron` to externals (#355)
- No longer need to 'ping' the main process to check the SDK has been enabled 

Disadvantages
- Requires a major version bump as it contains breaking changes
- The `session.setPreloads()` API was added in Electron v2 so that would become our minimum supported version (previously v1.8)
  - Electron v1.8.8 reached EOL 3 years ago
- ~~preload scripts have to be available on disk which presents some issues if the main process is bundled. Currently preload scripts get transpiled to variables in code and are written to `userData` directory at runtime~~
- If a user is already using the `session.setPreloads()` API, they will likely overwrite our preload scripts. I've added a link to the docs where we can help users diagnose and fix the issues they may encounter.


MarshallOfSound (changes requested):
From both Electrons perspective and Slacks this is unsafe.  We don't want JS written to disk and arbitrarily loaded without any form of validation, this bypasses gatekeeper and any other kind of signature validation.

IMO no tool or framework should make architectural decisions like this for apps by default, this kind of thing (if it had happened without us noticing) could cause untold security issues for downstream apps that the Sentry SDK might not consider but we definitely do.

> Users no longer have to supply appName in any process

You can solve this using `additionalArguments` in newly created webContents instead

> No longer need to 'ping' the main process to check the SDK has been enabled

Can you provide more context on why this ping is needed? I might have alternatives to remove it

timfish:
> IMO no tool or framework should make architectural decisions like this for apps by default, this kind of thing (if it had happened without us noticing) could cause untold security issues for downstream apps that the Sentry SDK might not consider but we definitely do.

I posted this PR to the Electron Discord security channel specifically because I doubted it was a sensible thing to do! 

> You can solve this using `additionalArguments` in newly created webContents instead

`appName` is only required in the renderer with Electron pre v9 because it's required when calling `crashReporter.start()`. 

I didn't know about `additionalArguments` and it looks like it's been there for years.

The problem with `additionalArguments` is that this SDK does not get involved in creating `BrowserWindow`s, and there's currently no way to intercept their creation (https://github.com/electron/electron/issues/13846). 

It's probably simpler to just require passing `appName` in the renderer if you're using older versions of Electron 🤷‍♂️.

> Can you provide more context on why this ping is needed? I might have alternatives to remove it

I think it is there mainly to improve developer experience. If the renderer cannot get an IPC response from the main process [it logs a `console.warn`](https://github.com/getsentry/sentry-electron/blob/0688d7dc783a4e79da0e6baeba088b10e0da69da/src/renderer/backend.ts#L198) reminding devs that you also need to `init` the SDK in the main process if you want anything to work.

timfish:
Closed in favour of #361

PR #364, opened 2021-08-31
Fix: Possible race condition with `initialScope` over IPC

timfish:
This change fixes a race condition over IPC found in the v3 branch where `initialScope` passed to the renderer could cause the initial scope to be lost in the main process.

`initialScope` was added so that we could have user ids in crashes sent via the build in electron uploader. That feature will only work if passed to `init` in the main process so this is a non-breaking change.

### DIFF
--- a/src/renderer/client.ts
+++ b/src/renderer/client.ts
@@ -13,6 +13,9 @@ export class RendererClient extends BaseClient<RendererBackend, ElectronOptions>
    * @param options Configuration options for this SDK.
    */
   public constructor(options: ElectronOptions) {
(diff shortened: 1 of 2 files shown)


src/renderer/client.ts
```diff
@@ -13,6 +13,9 @@ export class RendererClient extends BaseClient<RendererBackend, ElectronOptions>
    * @param options Configuration options for this SDK.
    */
   public constructor(options: ElectronOptions) {
+    // We only handle initialScope in the main process otherwise it can cause race conditions over IPC
+    delete options.initialScope;
```
AbhiPrasad, line 17: Isn't this a breaking change?

PR #373, opened 2021-10-17
feat: Test Recipes + Bug fixes

timfish:
This PR is a full rewrite of the e2e tests into self contained "recipes" which means the diff is quite large. 
- The full captured events/sessions are now checked against expected JSON
- Complex examples are now tested
- Recipe progress, app output and test server events are captured and output on test failure for easier debugging of failed tests

The good news is that this uncovered a number of issues:
- The `update-version.js` regex could not cope with `beta` versions!
- `release` could not be overridden for some errors/crash sources
- Some context was missing from some errors/crashes
- `@sentry/node` was picking up `release` from environment variables on CI!
- Global flag to stop `@sentry/browser` from being initialised more than once. This can occur if you `init` in the preload and renderer and have `contextIsolation` disabled
- Crashed sessions were not sent when using Electron uploader (because these don't touch the hub)


package.json
```diff
@@ -34,69 +34,68 @@
...
-    "electron": "15.0.0",
-    "electron-download": "^4.1.1",
-    "electron-mocha": "^6.0.4",
-    "extract-zip": "^1.6.7",
-    "koa": "^2.13.1",
+    "electron": "15.3.0",
+    "@electron/get": "^1.13.0",
+    "electron-mocha": "^10.1.0",
+    "extract-zip": "^2.0.1",
+    "koa": "^2.13.4",
```
AbhiPrasad, line 88: Can we possibly move these deps into a e2e test package.json? Or too much trouble you think?

src/main/integrations/electron-events.ts
```diff
@@ -75,7 +75,7 @@ export class ElectronEvents implements Integration {
         const breadcrumb = {
           category: 'electron',
           message: `${category}.${event}`,
-          timestamp: new Date().getTime() / 1000,
+          timestamp: new Date().getTime() / 1_000,
```
AbhiPrasad, line 78: Aren't numeric seperators Node 12+?

src/main/version.ts
```diff
@@ -1 +1 @@
-export const SDK_VERSION = '2.5.4';
+export const SDK_VERSION = '3.0.0-beta.2';
```
AbhiPrasad, line 1: Do we have to do this?

test/e2e/recipe/index.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    for (const event of testServer.events) {
+      event.data = normalize(event.data);
+    }
+
+    for (const [i, expectedEvent] of expectedEvents.entries()) {
+      delete expectedEvent.condition;
+
+      log(`Comparing event ${i + 1} of ${expectedEvents.length}`);
+      expect(testServer.events).to.containSubset([expectedEvent]);
```
AbhiPrasad, line 170: I was thinking it would be useful to leverage snapshot tests here somehow (so devs can automatically regenerate them if needed), but I guess that's harder to get working if we assert on the subset.

.github/workflows/build.yml
```diff
@@ -15,7 +15,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
```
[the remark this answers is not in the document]
timfish: I was messing around with some examples that used modules (ie. `.mjs`) and eslint complained. No idea why this needs to be explicitly enabled 🤷

src/main/integrations/electron-events.ts
timfish: Something like that, however we're using TypeScript targeting ES6 so these get transpiled down to ES6 compatible JavaScript!

test/e2e/recipe/index.ts
timfish: I considered snapshots but the json events are modified/cleaned up to make the tests more reliable across Electron versions. For example, for most of the tests, I've stripped all the breadcrumbs out because these vary depending on the events supported for each Electron version.

src/main/version.ts
timfish: I should actually remove this file from source control, add it to `.gitignore` and generate it on every build!

.github/workflows/build.yml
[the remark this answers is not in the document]
timfish: The default for `setup-node@v1` was Node v10 and after updating all the dependencies some no longer supported v10:
https://github.com/getsentry/sentry-electron/runs/4062090981?check_suite_focus=true#step:5:9

package.json
timfish: Yes, it might be nice to split it into a workspace with SDK + Tests.

Shall I do that in another PR?

AbhiPrasad: Yup separate PR sounds like the best approach

Issue #335, opened 2021-05-27
Submit events for `child-process-gone` events


timfish:
Electron >= v11 has the [`child-process-gone`](https://www.electronjs.org/docs/api/app#event-child-process-gone) event. It would be useful to optionally generate sentry events for this.

PR #375, opened 2021-10-28
feat: add error reporting for electron child-process-gone

JCMais:
Closes https://github.com/getsentry/sentry-electron/issues/335

I'm not sure how to create a test for this, so any help is appreciated. 

For electron versions older than 11 I'm reporting errors just for `gpu-process-crashed`.


timfish:
Calling `this._sendNativeCrashes(options, event)` will only send events to Sentry if a minidump file is subsequently found.

The `child-process-gone` event is triggered for the following processes:
`Utility, Zygote, Sandbox helper, GPU, Pepper Plugin, Pepper Plugin Broker, Unknown`

And you only attempt to send for:
`abnormal-exit, crashed, oom, launch-failed, integrity-failure`

I know for sure that `GPU`+`Crashed` with Crashpad will result in a minidump but I'm not sure which others will actually result in a minidump.

Issue #372, opened 2021-10-11
V3 Beta - errors not reported from preload context


timfish:
## Versions + Platform

- [x] SDK version - `@sentry/electron@v3.0.0-beta.1`
- [x] Electron version - `all`
- [x] Platform - `all`

## Description

With v3 beta, errors are not reported from the preload context.

timfish:
I'm nowhere near a machine to test this but you can probably work around this by initialising the render code in the preload too:

```
import '@sentry/electron/preload';
Import { init } '@sentry/electron/renderer';

init({});
```

If you're using context isolation, you will need to init the renderer code in the preload AND the isolated renderer context.

timfish:
I think reporting errors from the preload context should work out of the box by default but this adds some annoying complications that vary by the users Electron configuration.

For non-bundled apps, we automatically inject the preload script which is currently well under 1kB in size. If we want to support reporting errors from this context with zero config, we'll need to include all the renderer code which is currently ~78kB. If we include this code by default, we'll also need to call `init` and users will not be able to change the settings passed to `init`. In the preload context, passing options to `init` might not be a huge issue except...

If users have `contextIsolation` disabled (not the default and not recommended, but very common), the preload and renderer context are not distinct so once the SDK has been `init`'ed  in the preload, it's effectively setup in the renderer and there is no way to pass option to `init`. 

Why not let users just create their own preload script and choose if they want to support error reporting from this context and pass custom `init` options? Well, if `sandbox` mode is enabled (planned to become default), the preload has no access to `require`, so preload scripts will need to be bundled!

## To summarise

- If we `init` the renderer SDK in the preload, users with `contextIsolation: false` cannot pass options to `init`
- If we don't `init` the renderer SDK in the preload, users that want errors from preload context will need to use a bundler if `sandbox: true`

timfish:
This is simpler in 3.0.0-beta.2

A preload is no longer required unless you're using Electron < v5 and bundling the main process.

If you would like errors reported from the preload then you'll still need to initialise the renderer portion of the sdk there. No options are required unless you need to configure specific renderer integrations.

`preload.js`
```ts
const { init } = require('@sentry/electron/renderer');
init();
```

If you have context isolation enabled, the renderer is a different context to the preload so you'll need to `init` the sdk again there.

its-tim-lee:
Hey, @timfish 
[...]

PR #377, opened 2021-11-04
feat: Fallback to custom protocol

### LINKED ISSUE #376: Proposal: (almost) Entirely remove the need for preload script
## Problem

Even after the large refactor for v3, the Electron SDK can be a little at odds with Sentry's [developer philosophy](https://develop.sentry.dev/sdk/philosophy/):
- We optimize towards a great out of the box experience
- Assume Novice Developers 
- etc, etc

Electron is a complex beast.

Electrons IPC modules are the primary means to communicate between the main and renderer processes. However the tightening of the security model and the many configuration options has made it complex for library authors and their consumers. The only reliable way to expose IPC across all configurations is via `contextBridge` in a preload script. We can make a best effort at injecting preload scripts but this is not compatible with bundlers and instructing users to configure preload scripts with various bundlers is not simple and still leaves outstanding issues (https://github.com/getsentry/sentry-electron/issues/372#issuecomment-955803319) unsolved.

There is another option! 

The [`protocol`](https://www.electronjs.org/docs/latest/api/protocol) module allows you to setup a custom protocol from the main process which you can call from a renderer via `fetch`. 

This has a few downsides over IPC:
- Only fully supported with CSP in Electron >= v5
- No easy way to initiate main to renderer communication
  - Not a problem for us as we don't do this
  - Turns out you can emulate this by [keeping stream open](https://github.com/timfish/electron-protocol-ipc/blob/4e3f0bf9aeff6c65374939e4f7a2540a7f52cc90/src/ipc/renderer.ts#L42-L59)!
- No way to tell which renderer made the call
  - Less context over which renderer the error came from

## Proposal

- Setup IPC in the main process
- Setup `sentry-ipc` protocol in main process when Electron >= v5
- Attempt to inject preload scripts
- In renderers, favour IPC but fallback to `fetch('sentry-ipc://event-etc')`

### Unsupported Configurations

When using Electron < v5 while bundling the main process code there is no custom protocol and preload injection will fail. Users will be required to configure a preload script. 

### Downsides

When users bundle the main process code, the SDK cannot inject preload scripts, will fallback to `fetch` + custom protocol and you'll loose context over exactly which renderer JavaScript errors come from. The URL/path information is all sill included. 

Users can negate this by configuring a preload script.

### Upsides

Unless you're using Electron < v5 and bundling your main process code, there is no longer a need to care about preload scripts.

timfish:
Closes #376 and #372

Adds support for receiving renderer events and scope via a custom `protocol`.

- Add `ipcMode: IPCMode` option to main process. Default `IPCMode.Both` enables both IPC and protocol for best compatibility
- Add `getSessions: () => [session]` option to main process.  Defaults to `() => [session.defaultSession]`
  - This allows users to pass in custom `session`s if `defaultSession` doesn't suffice
  - We use `session`s for setting up the protocol and preload injection
  - This option is a function because `init` must be called before app `ready` and `session`s cannot be accessed until after app `ready`!
- Configures custom protocol in main process when Electron >= v5 and when `ipcMode` is `IPCMode.Protocol` or `IPCMode.Both`
- Renderers send events and scope via IPC or protocol. IPC is favoured because it has a bit more context
- `PreloadInjection` integration modified to use `getSessions` option and not to run at all when `ipcMode` is `IPCMode.Protocol`
- Modified preload scripts to guard from being run multiple times
  - This could occur if preload injection succeeds and user also adds preload 
- Modified existing examples to no longer use preload

Also:
- Moves `contexts.electron.crashed_process` to `tags['event.process']` so users can filter by process type

Added Tests
- Webpack example WITH preload
- Renderer error with `ipcMode: IPCMode.Protocol`
- Preload JavaScript error 
- Preload JavaScript error with `ipcMode: IPCMode.Protocol`
- Webpack test using protocol with strict `Content-Security-Policy`
- Sending JavaScript error from `<iframe>` within renderer!
- Console log when SDK not started in main process
- Throws error when `init` after app ready


timfish:
I haven't yet decided what to do with `processName` and/or `getRendererName`. 

🤔

timfish:
Currently v2 and v3-beta add the following contexts:

Renderer JavaScript and native crashes
```json
    "contexts": {
      "electron": {
        "crashed_process": "WebContents[1]",
        "crashed_url": "app:///src/index.html"
      }
    },
```

Main process JavaScript and native crashes:
(Confusingly and forever stuck for backwards compatibility, Electrons main process is called `browser`)
```json
    "contexts": {
      "electron": {
        "crashed_process": "browser"
      }
    },
```

The naming of these `electron.crashed_*` properties suggests that they should only be populated for native crashes which is currently not the case. In addition, having these in `contexts` means users will not be able to filter on them.

I would suggest that we move `contexts.electron.crashed_process` to a tag and set it for any kind of event. This means we'd end up with:

| Tag | Possible Values |
|---|---|
| `event.origin` | `electron` |
| `event.environment` | `javascript` \| `native` |
| `event.process` | `browser` \| `renderer[id]` \| result of `getRendererName()` |

Or would `event.origin` be a more suitable tag for the source process? Open to suggestions!

When it comes to URLs, we already have these for renderer JavaScript events via `request.url`. For renderer native crashes, shall we keep this at `contexts.electron.crashed_url`?

src/renderer/ipc.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  fetch(`${PROTOCOL_SCHEME}://${IPCChannel.PING}`).catch(() =>
+    console.error(`Sentry SDK failed to establish connection with the Electron main process.
+ - Ensure you have initialized the SDK in the main process
+ - If your renderers use custom sessions, be sure to set 'getSessions' in the main process options
+ - If you are bundling your main process code and using Electron < v5, you'll need to manually configure a preload script`),
+  );
+
+  // We include sentry_key in the URL so these dont end up in fetch breadcrumbs
+  // https://github.com/getsentry/sentry-javascript/blob/a3f70d8869121183bec037571a3ee78efaf26b0b/packages/browser/src/integrations/breadcrumbs.ts#L240
```
AbhiPrasad, line 24: This is kinda hacky - but a nice solution

AbhiPrasad (commented):
Nice, the tests make everything pretty easy to understand.

> Won't work if the renderer crashes before SDK init

I think we shouldn't spend too much time on this. Other platforms have a similar problem - the only solution is if the framework/OS gives us a way to read those crash events afterwards.

> If multiple renderers use the same path, we can only link them by the order that they are started which will not be reliable

Is it a common pattern in Electron apps for multiple renderers to use the same path? If so, let's just not worry too much about linking `processName` and `window.location.href`, maybe we leave it for a future patch and just open an issue to collect feedback.

> would event.origin be a more suitable tag for the source process

IMO the 3 tags shown in the table provide adequate cardinality to identify the crash location, don't think we need to change `event.origin`.

> For renderer native crashes, shall we keep this at contexts.electron.crashed_url

Yup think this is a great idea.

Lmk and I can go over for a final review. Thanks Tim, you a champ!

src/renderer/ipc.ts
timfish: Yeah, maybe a bit fragile. I initially attempted to filter the breadcrumbs when they're sent from renderer to main but realised that scope updates were triggering fetch, which was triggering scope updates, which.... yeah not great!

Might be nice to have a magic way to disable Sentry fetch breadcrumbs via a magic header?

AbhiPrasad: Lemme ask around to see what others think.

timfish:
> Is it a common pattern in Electron apps for multiple renderers to use the same path?

I imagine so.  

We've got a production app that has multiple windows that use a single entry point and the UI varies entirely by the startup hash/fragment identifier . If your multiple windows share most of their code, this is much easier than configuring webpack for shared dependency chunks between multiple entry points!

Anyway, my hairbrained idea was linking renderers via the `fetch` `referrer` URL which I've since remembered doesn't include the fragment identifier so that wouldn't have been be very useful 😂

src/renderer/ipc.ts
timfish: We're pinned to a specific version of `@sentry/browser` so this is only prone to breaking on dependency updates. Shall we merge this as it is and I'll open an issue on `getsentry/javascript` to track this?

AbhiPrasad: Yes please, opening an issue would be great

PR #379, opened 2021-11-08
Fix: Handle previously crashed/abnormal sessions

timfish:
3.0.0-beta.1 brought release health via sessions but this failed to cater for native crashes in the main process or abnormal exit.

This PR:
- Refactors session handling to its own file outside the current default integration
- Stores the active session in `session.json` on session start and updates every minute
- Clears the active session file on clean exit
- On startup if  `session.json` exists:
  - With Sentry uploader, report previous session as `crashed` if any minidumps were found, otherwise `abnormal`.
  - With Electron uploader, if `getLastCrashReport()` suggests that a crash occurred between last session persist and estimated end of session report as `crashed`, otherwise `abnormal`

Tests:
- Runs `Good Session` test twice to ensure no abnormal session for first run
- Adds test for crashed main process Sentry uploader
- Adds test for crashed main process Electron uploader
- Adds tests for abnormal exit Sentry uploader
- Adds tests for abnormal exit Electron uploader


src/main/sessions.ts
```diff
@@ -0,0 +1,88 @@
...
+    logger.log(`Found previous ${status} session`);
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    const transport = (getCurrentHub().getClient<NodeClient>() as any)
+      ._getBackend()
+      .getTransport() as ElectronNetTransport;
+
+    previousSession.status = status;
+    previousSession.errors = (previousSession.errors || 0) + 1;
+
+    await transport.sendSession(previousSession);
```
AbhiPrasad, line 63: Would rather just construct a new session with

```suggestion
    await transport.sendSession({ ...previousSession, status, errors: (previousSession.errors || 0) + 1 });
```

and not have to worry about mutability issues from using `let previousSession`, but I'll leave it to you to make the final call.

src/main/store.ts
```diff
@@ -36,14 +36,18 @@ export class Store<T> {
    * Updates data by replacing it with the given value.
    * @param next New data to replace the previous one.
    */
-  public set(next: T): void {
+  public set(next: T, forceFlush: boolean = false): void {
```
AbhiPrasad, line 39: need to update doc string for this method.

AbhiPrasad (approved):
Nice, I like moving the logic into `sessions.ts`.

src/main/sessions.ts
timfish: `previousSession` is mutable solely so it can be set as `undefined` once its been sent to stop it being sent again.  `const previousSession` wouldn't stop any of its properties from being modified

src/main/sessions.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/** Checks if the previous session needs sending as crashed or abnormal  */
+export async function checkPreviousSession(crashed: boolean): Promise<void> {
+  if (previousSession) {
+    const status = crashed ? SessionStatus.Crashed : SessionStatus.Abnormal;
+
+    logger.log(`Found previous ${status} session`);
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    const transport = (getCurrentHub().getClient<NodeClient>() as any)
```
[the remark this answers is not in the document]
timfish: `client.captureSession()` requires a proper `Session` object because it calls `session.update({ init: false })` but `previousSession` from json file is effectively only a serialised `SessionContext` interface. 

Probably not worth creating a `Session` which would be immediately discarded?

We do already use `unsafe-member-access` and `_getBackend` in the minidump uploader to get the transport.
🤔

test/e2e/test-apps/sessions/abnormal-exit/src/main.js
```diff
@@ -0,0 +1,34 @@
...
+  debug: true,
+  onFatalError: () => {},
+});
+
+app.on('ready', () => {
+  const mainWindow = new BrowserWindow({
+    show: false,
+    webPreferences: {
+      nodeIntegration: true,
+      contextIsolation: false,
```
iker-barriocanal, line 17: Should this be configurable?

iker-barriocanal (commented):
I'm not familiar with electron, so haven't looked much into the code and going to throw some thoughts/questions first:

- The scope of the PR could be smaller.
- Why is storing `session.json` required? If this is not trivial with electron knowledge, we should write the reason for going with this approach in the PR description.
- What format does `session.json` expect? Are all fields required? Are there optional fields? What is it for? Is it required, or is it optional?
- I also see a `session-previous.json`, and I have the same questions as I have with `session.json`.
- Why is `Good Session` required to be run twice to ensure "no abnormal session report"? What does an "abnormal session report" mean? Is it documented somewhere, or should we clarify what it means? I see it has been refactored from `session-native-crash`, are they the same? What's the difference? Why this refactor? What implications does this refactor have (e.g. does the data on Sentry going to be different for users?)?
- If running twice is required, can we make it so it always runs twice (or the number of times its required), instead of relying on someone running it that many times? For example, if I had to contribute to this repo and had no idea about this, how could I prevent flaky tests or "abnormal session report"s?

timfish:
@iker-barriocanal thanks for the feedback!

My main take-away from your review is that how the integration tests work is non-obvious and needs documenting!

Here's a quick recap:
- Everything under `/test/e2e/test-apps/*` and `/examples*` are self contained Electron apps used for integration tests.
- Each test contains a `README.md` with a title and some options in a table which is pared [here](https://github.com/getsentry/sentry-electron/blob/master/test/e2e/recipe/parser.ts)
- Each test is prepared and run via the [`RecipeRunner`](https://github.com/getsentry/sentry-electron/blob/master/test/e2e/recipe/index.ts)
- We start a server which receives all events/sessions from the test app
- Some [normalization](https://github.com/getsentry/sentry-electron/blob/master/test/e2e/recipe/normalize.ts) is applied to received events so they can be compared
- Any `event*.json` or `session*.json` files with a test app are used for comparing expected events and confirming if the tests pass or fail using `chai-subset`.  

So `test/e2e/test-apps/sessions/abnormal-exit/session.json` and `test/e2e/test-apps/sessions/abnormal-exit/session-previous.json` are two envelopes that we expect to receive from the `abnormal-exit` test app.

> Why is storing session.json required? If this is not trivial with electron knowledge, we should write the reason for going with this approach in the PR description.

The Sentry developer docs about [abnormal sessions](https://develop.sentry.dev/sdk/sessions/#crashed-abnormal-vs-errored) suggest this kind of approach is common/expected:

> Abnormal session ends are normally to be recorded on application restart.

> These are situations where the user forced the app to close via the task manager, the machine lost power or other situations. A session can be stored by persisting it to disk eagerly. This saved file can be detected on application restart to close the session as abnormal.

In the PR description when I refer to `session.json`, that's a file saved at runtime by the `Store` class and the presence of that file is detected at app startup and used to determine if the previous session exited cleanly or not.

> Why is Good Session required to be run twice

Some of the integration test apps are required to run twice because the events or sessions cannot be sent before the app terminates. 

We can only know if a session exited abnormally on the next app run. Consequently, we can only know that a "good session" isn't later incorrectly marked as abnormal if we run the app for a second time to test this.

> If running twice is required, can we make it so it always runs twice

[Any integration test](https://github.com/getsentry/sentry-electron/blob/master/test/e2e/test-apps/native-sentry/main/README.md?plain=1#L7) that defines `Run twice: true" in it's `README.md` table, will automatically be run twice.

@AbhiPrasad [suggested](https://github.com/getsentry/sentry-electron/pull/373#discussion_r740987879) refactoring this repository into a workspace to isolate the tests from the SDK. I still intend to do this but since that will make merging any existing code a total pain, I'm working through some of my other existing changes first. When I do this refactor, I'll add some detailed test documentation!

test/e2e/test-apps/sessions/abnormal-exit/src/main.js
timfish: This is an integration test app. Most of the tests are run with `nodeIntegration: true, contextIsolation: false` because otherwise we need to bundle the renderer code due to lack of `require`.

src/main/sessions.ts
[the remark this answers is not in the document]
AbhiPrasad: Ah yeah - we should stick with the transport.

src/main/sessions.ts
```diff
@@ -0,0 +1,138 @@
...
+import { flush, NodeClient } from '@sentry/node';
+import { SessionContext, SessionStatus } from '@sentry/types';
+import { logger } from '@sentry/utils';
+import { app } from 'electron';
+import { join } from 'path';
+
+import { Store } from './store';
+import { ElectronNetTransport } from './transports/electron-net';
+
+const PERSIST_INTERVAL = 60_000;
```
iker-barriocanal, line 11: What's the unit of this number?

timfish:
@AbhiPrasad probably worth one last look.

After this is merged I'll do a 3.0.0-beta.2 release and then work on re-org into workspace with added test documentation.

src/main/sessions.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  const previousSessionModifiedTime = previousSessionModified.getTime();
+  const crashTime = crashDate.getTime();
+
+  // Session could have run until modified time + persist interval
+  const prevSessionEnd = previousSessionModifiedTime + PERSIST_INTERVAL;
+
+  // Event cannot be much before last persist time
+  const lastPersist = previousSessionModifiedTime - 2_000;
```
iker-barriocanal, line 85: Where is `2_000` coming from? What does it mean?

iker-barriocanal (approved):
LGTM. Considering that some `.md` files are used as config files, it could be better to have more config-friendly files such as `yaml`.

timfish:
> could be better to have more config-friendly files such as yaml

Agreed. I used markdown so we could have additional user docs with examples but there's nothing stopping me from putting a yaml header in markdown, which GitHub handily renders like [this](https://gist.github.com/timfish/37d409151808b9ba63ffbade212bace1).

```md
---
somekey: some value 
anotherKey: more here
---

# Test title

More docs here
```

iker-barriocanal:
Sure, this can be addressed in the future. Good job!

AbhiPrasad (approved):

AbhiPrasad:
@timfish One thing that I've been thinking about (and please shoot me down if it's not a problem), is how two instances of the same electron app interact with the stored sessions. Is this even a problem? Are there race conditions on reading/writing to disk?

Also (these functions don't exist in JS but we can add em), do you think it'll be valuable to pause/resume sessions based on if they are in the foreground/background? Mobile does something similar. We could even end the session when app goes to background and start a new one on foreground instead of pausing/resuming.

timfish:
> One thing that I've been thinking about (and please shoot me down if it's not a problem), is how two instances of the same electron app interact with the stored sessions. Is this even a problem?

Although Electron does support multiple instances, it's [not supported or tested](https://github.com/electron/electron/issues/4727) for more than one instance to use the same `userData` directory. There are loads of things that break if you try, for example, [`localStorage` and `indexedDb` just fail to work](https://github.com/electron/electron/issues/2493) because the underlying `LevelDb` only support single process. 

It's recommended that you only ever run a single main process and use multiple renderers to create what feels like a  "multi-process" application. If a 2nd instance starts, it signals its arguments to the 1st instance and closes immediately [like this](https://www.electronjs.org/docs/latest/api/app#apprequestsingleinstancelock). Chrome does something similar and Electron apps have to follow suit.

Since we save `session.json` under `userData`, we don't have to worry about multiple processes accessing the files.

> We could even end the session when app goes to background and start a new one on foreground instead of pausing/resuming

The current `MainProcessSession` integration only covers the most likely session scenario for Electron apps. We could add a `BrowserWindowSession` that tracks every window as a separate session and aggregate them in the main process. That integration could have an option to track sessions via visibility rather than the full process lifetime. We could even allow users to have both integrations at the same time although we'd need some way to annotate them. Can Sentry deal with multiple session types that are split out in the UI?

AbhiPrasad:
> We could even allow users to have both integrations at the same time although we'd need some way to annotate them. Can Sentry deal with multiple session types that are split out in the UI?

Bringing this up with the larger team (as well as seeing if we can define a spec for desktop sessions in general). Should have an answer by end of next week.

PR #381, opened 2021-11-22
test: Electron v16

timfish:


timfish:
Closing in favour of #383 which automatically fetches the latest Electron versions

PR #383, opened 2021-11-27
test: Improve tests and documentation

timfish:
- Migrates test metadata from funky table markdown to yaml files
- Adds test documentation

In CI every day, we check if the list of latest versions and if they change, open or update a PR with this change.


package.json
```diff
@@ -44,6 +44,7 @@
     "build:watch:esm": "tsc -p tsconfig.esm.json -w --preserveWatchOutput",
     "clean": "rimraf coverage esm main preload renderer index.* integrations.* ipc.* sentry-electron*.tgz",
     "prelint": "node scripts/update-version.js",
+    "electron-versions": "electron-latest-versions --start 2",
```
iker-barriocanal, line 47: By looking at the [docs](https://github.com/timfish/electron-latest-versions#cli), major versions are included and may be returned by running this command. I think it's okay to run tests in new patches or even minors, but I don't think we should break CI for tests failing in a new major release. What do you think? Is there something I'm missing?

iker-barriocanal (commented):
I'm not sure if we want to run weekly tests on new versions. Should CI be red if the SDK doesn't support the latest version? Same doubts as in a comment below with using different release versions.

What is the motivation behind this feature? What problem is it trying to solve?

timfish:
We could fix the tested versions: `electron-latest-versions --start 2 --end 16`, but I wanted a way to automatically test new Electron releases and be notified if the tests fail. Electrons features are now relatively stable and I don't expect new versions to be incompatible with the SDK, but they'll still need testing in a timely manner.

The ideal way to do this would be to:
- Periodically check for new major Electron versions
- When a new version is found, update `--end {version}` and open a PR
- Keep the PR updated with new minor/patch releases

I'll see how easy it is to achieve this!

timfish:
New approach:

Every day in CI we check if the latest versions have changed and create/update a PR with these changes.

AbhiPrasad (approved):
awesome

timfish:
Now we just have to wait and see if the automatic PR generation works!

iker-barriocanal:
🚀

PR #388, opened 2021-11-30
fix: Actually make CORS requests

timfish:
Turns out enabling CORS for the protocol was not enough. We need to actually make CORS mode fetch requests.

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -86,7 +86,7 @@ function configureProtocol(options: ElectronMainOptions): void {
             handleScope(options, data);
           }
 
-          callback('');
+          callback({ data: '', headers: { 'Access-Control-Allow-Origin': '*' } });
         });
       }
     }),
--- a/src/renderer/ipc.ts
+++ b/src/renderer/ipc.ts
@@ -13,7 +13,7 @@ function getImplementation(): IPCInterface {
 
   logger.log('IPC was not configured in preload script, falling back to custom protocol and fetch');
 
-  fetch(`${PROTOCOL_SCHEME}://${IPCChannel.PING}`).catch(() =>
+  fetch(`${PROTOCOL_SCHEME}://${IPCChannel.PING}`, { mode: 'cors' }).catch(() =>
     console.error(`Sentry SDK failed to establish connection with the Electron main process.
  - Ensure you have initialized the SDK in the main process
  - If your renderers use custom sessions, be sure to set 'getSessions' in the main process options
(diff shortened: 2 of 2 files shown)


src/renderer/ipc.ts
```diff
@@ -24,12 +24,14 @@ function getImplementation(): IPCInterface {
...
-      fetch(`${PROTOCOL_SCHEME}://${IPCChannel.SCOPE}/sentry_key`, { method: 'POST', body: scope }).catch(() => {
+      const init: RequestInit = { mode: 'cors', method: 'POST', body: scope };
+      fetch(`${PROTOCOL_SCHEME}://${IPCChannel.SCOPE}/sentry_key`, init).catch(() => {
         // ignore
       });
     },
     sendEvent: (event) => {
-      fetch(`${PROTOCOL_SCHEME}://${IPCChannel.EVENT}/sentry_key`, { method: 'POST', body: event }).catch(() => {
+      const init: RequestInit = { mode: 'cors', method: 'POST', body: event };
+      fetch(`${PROTOCOL_SCHEME}://${IPCChannel.EVENT}/sentry_key`, init).catch(() => {
```
AbhiPrasad, line 34: ah the `sentry_key` hack lol. The team is still undecided on how to best resolve that, so it'll stay like this for a while.

As an aside point, we should probably have a quick e2e test to regression test against adding ipc fetch breadcrumbs - just in case we mess something up in the main JS SDK. We should do it outside of this PR though.

src/main/ipc.ts
```diff
@@ -86,7 +86,7 @@ function configureProtocol(options: ElectronMainOptions): void {
             handleScope(options, data);
           }
 
-          callback('');
+          callback({ data: '', headers: { 'Access-Control-Allow-Origin': '*' } });
```
AbhiPrasad, line 89: Should we specifically allowlist the fetch urls we use?

AbhiPrasad (approved):
Would be nice to test this, but I can see how it can be tricky.

src/renderer/ipc.ts
timfish: I'll give it a test but I seem to remember that without this, most of the tests fail because we end up in an infinite loop creating breadcrumbs and sending scope that results in more breadcrumbs!

timfish:
> Would be nice to test this

I'll see what I can do!

src/main/ipc.ts
timfish: We could 404 for unknown URLs 🤔

AbhiPrasad: Fine with adding `404` in another PR. We can probably just debug log it out.

timfish:
This is blocked until I get a way to reproduce the CORS issues reported.

PR #390, opened 2021-12-01
feat: Additional context and breadcrumbs

### LINKED ISSUE #386: Instrument `BrowserWindow` events
We already instrument some useful `WebContents` events but `BrowserWindow` has events for window states that would be useful to have in breadcrumbs.

timfish:
This PR:
- Refactors and simplifies `src/main/context.ts` a little
    - Adds `context.app.app_start_time` 
- Renames `ElectronEvents` integration to `ElectronBreadcrumbs`. We should use Sentry terminology to avoid confusion.
    - Makes all the breadcrumb categories optional and enabled by default
    - Allows fine-grained configuration over which events need breadcrumbs
    - Adds breadcrumbs for `BrowserWindow` and `autoUpdater` events
    - For `BrowserWindow` and `WebContents` events, we now include the `id` and window title as extra breadcrumb data
- Adds default enabled `AdditionalContext` integration with all options enabled by default `{ cpu, screen, memory, language }`
    -  `cpu` adds `processor_count`, `cpu_description` and `processor_frequency` to `context.device`
        - Adds `context.device.machine_arch = 'arm64'` if machine arch is `arm64` and running in emulation 
    - `screen` adds `screen_resolution` and `screen_density` to `context.device`
    - `memory` adds `memory_size` and `free_memory` to `context.device`
    - `language` adds `language` to `context.device`
- Updates the test events and test normalization to test much of the above

All context in `src/main/context.ts` is not user configurable. The `AdditionalContext` integration allow this extra context to be configurable. The only downside to using an integration is that anyone using the `ElectronMinidump` uploader rather than the default sentry uploader will not get this additional context because those events don't go through the event processors.

Closes #386 

## `ElectronBreadcrumbs` configuration

Stop getting `app` events:
```ts
new ElectronBreadcrumbs({ app: false })
```

Only get `show` and `hide` events for `BrowserWindow`
```ts
new ElectronBreadcrumbs({ browserWindow: ['show', 'hide'] })
```

Get all `app` events apart from those starting with `remote-`:
```ts
new ElectronBreadcrumbs({ app: (name) => !name.startsWith('remote-') })
```


src/main/integrations/electron-breadcrumbs.ts
```diff
@@ -0,0 +1,221 @@
...
+
+          const webContentsName = initOptions?.getRendererName?.(contents) || 'renderer';
+
+          if (this._options.webContents) {
+            this._instrumentBreadcrumbs(webContentsName, contents, (name) => this._options.webContents?.(name));
+          }
+
+          if (this._options.unresponsive != false) {
+            contents.on('unresponsive', () => {
+              captureMessage(`${webContentsName} Unresponsive`);
```
AbhiPrasad, line 180: should we give this a `error` (or above) Severity level?

test/e2e/recipe/normalize.ts
```diff
@@ -75,6 +80,38 @@ function normalizeEvent(event: Event): Event {
...
+  if (event.contexts?.device?.memory_size) {
+    event.contexts.device.memory_size = 0;
+  }
+
+  if (event.contexts?.device?.free_memory) {
+    event.contexts.device.free_memory = 0;
+  }
+
+  if (event.contexts?.device?.processor_count) {
+    event.contexts.device.processor_count = 0;
```
AbhiPrasad, line 92: Would be nice to have unit tests to sanity check this integration. Perhaps we can make a big TODO issue and open it up as a `good first issue` for the community (or a future Sentry intern 😄)

AbhiPrasad (commented):
1st pass - exciting!!

src/main/integrations/additional-context.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  /** @inheritDoc */
+  public setupOnce(addGlobalEventProcessor: (callback: EventProcessor) => void): void {
+    addGlobalEventProcessor(async (event: Event) => this._addAdditionalContext(event));
+
+    // Some metrics are only available after app ready so we lazily load them
+    void whenAppReady.then(() => {
+      const { language, screen } = this._options;
+
+      if (language != false) {
```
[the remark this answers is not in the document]
timfish: The default of `undefined` is enabled. To disable you have to specifically set to false.

src/main/integrations/electron-breadcrumbs.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  private readonly _options: ElectronBreadcrumbsOptions<EventFunction | undefined>;
+
+  /**
+   * @param _options Integration options
+   */
+  public constructor(options: Partial<ElectronBreadcrumbsOptions<EventTypes>> = {}) {
+    this._options = {
+      unresponsive: options.unresponsive != false,
+      app: getOptionOrDefault(options, 'app'),
```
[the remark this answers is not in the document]
timfish: I could remove the `unresponsive: boolean` property and then iterate through `Object.keys`

src/main/integrations/additional-context.ts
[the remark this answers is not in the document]
AbhiPrasad: Maybe we can set the default to be `true` then. Something like:

```js
export interface AdditionalContextOptions {
  cpu: boolean;
  screen: boolean;
  memory: boolean;
  language: boolean;
}

const DEFAULT_OPTIONS = {
  cpu: false,
  screen: false,
  memory: false,
  language: false,
}

// In the class
public constructor(options: Partial<AdditionalContextOptions> = {}) {
  this._options = {
    ...options,
    ...DEFAULT_OPTIONS,
  }
}
```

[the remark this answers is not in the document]
timfish: Ah, that's a good point. If we convert all `true` to `undefined`, spread will fill all `undefined` with defaults...

src/main/integrations/electron-breadcrumbs.ts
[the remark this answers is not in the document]
timfish: If I convert all the `true` to `undefined` I should then be able to spread the defaults over the user supplied options 🤔...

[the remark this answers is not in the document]
AbhiPrasad: Yeah I think that works well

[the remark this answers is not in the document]
timfish: Can you think of a nicer (typesafe) way to do it than this?
```ts
    (Object.keys(options) as (keyof ElectronBreadcrumbsOptions<EventTypes>)[]).map((k) => {
      if (options[k] === true) {
        options[k] = undefined;
      }
    });
```

src/main/context.ts
```diff
@@ -319,9 +326,9 @@ export async function getEventDefaults(release?: string): Promise<Event> {
...
   // promise here synchronously to avoid multiple events computing them at the
   // same time.
-  if (!defaultsPromise) {
-    defaultsPromise = _getEventDefaults(release);
+  if (!cachedDefaultsPromise) {
+    cachedDefaultsPromise = _getEventDefaults(release);
   }
 
-  return await defaultsPromise;
+  return await cachedDefaultsPromise;
```
AbhiPrasad, line 333: I wonder if we should cache the actual value instead of having to resolve the promise every time.

src/main/integrations/electron-breadcrumbs.ts
```diff
@@ -0,0 +1,206 @@
...
+  public name: string = ElectronBreadcrumbs.id;
+
+  private readonly _options: ElectronBreadcrumbsOptions<EventFunction | undefined>;
+
+  /**
+   * @param _options Integration options
+   */
+  public constructor(options: Partial<ElectronBreadcrumbsOptions<EventTypes>> = {}) {
+    // Convert all 'true' properties to undefined since undefined uses the defaults too
+    (Object.keys(options) as (keyof ElectronBreadcrumbsOptions<EventTypes>)[]).map((k) => {
```
AbhiPrasad, line 99: We shouldn't use `.map` if we aren't gonna use the return value, `.forEach` probably applies better here.

Mutating the arg is not ideal though.

Think there's no way around the casting of the key due to: https://github.com/microsoft/TypeScript/pull/12253#issuecomment-263132208. `Object.keys` will probably never go generic :/

test/e2e/recipe/normalize.ts
timfish: We could certainly have better normalisation of events. For example, check that everything we normalise is at least reasonable and close to what's expected.

`chai-subset` makes the event comparisons 1 line but we could certainly have something better!

AbhiPrasad: Can we have assertions similar to `expect.any(String)` or `expect.any(Number)` before we apply the normalization to check? Not convinced its the best idea though.

src/main/integrations/electron-breadcrumbs.ts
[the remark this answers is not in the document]
AbhiPrasad: > Can you think of a nicer (typesafe) way to do it than this?

Replied in: https://github.com/getsentry/sentry-electron/pull/390#discussion_r762156383

src/main/integrations/electron-breadcrumbs.ts
AbhiPrasad: Opened https://github.com/getsentry/sentry-javascript/pull/4229 to lint against the `.map` stuff.

timfish: ah, and this doesn't handle `string[]` inputs either

src/main/integrations/electron-breadcrumbs.ts
timfish: I've decided to pull `unresponsive` out of this integration as it doesn't really fix in here.

I'll create a new integration in another PR to address this and also other Electron events where you might want to generate Sentry rather than breadcrumbs (ie. #335)

test/e2e/recipe/normalize.ts
timfish: We need an object comparison/diffing library that lets us put ranges or custom validation functions...

```json
  "timestamp": "validate:timestamp",
```

timfish: It looks like `chai-subset` already supports functions in the expected event:
https://github.com/debitoor/chai-subset/blob/5c943ada7356c23d3bd9b04ba2b230ec8a5eebac/lib/chai-subset.js#L74-L76

Just need to create a JSON revive function that replaces validation strings with the correct functions!

src/main/integrations/electron-breadcrumbs.ts
```diff
@@ -0,0 +1,198 @@
...
+  public name: string = ElectronBreadcrumbs.id;
+
+  private readonly _options: ElectronBreadcrumbsOptions<EventFunction | undefined>;
+
+  /**
+   * @param _options Integration options
+   */
+  public constructor(options: Partial<ElectronBreadcrumbsOptions<EventTypes>> = {}) {
+    this._options = {
+      ...normalizeOptions(options),
```
AbhiPrasad, line 107: Would be nice to unit test this

AbhiPrasad (approved):

src/main/integrations/electron-breadcrumbs.ts
timfish: Added!

PR #392, opened 2021-12-02
feat: Offline Support

### LINKED ISSUE #369: Offline integration does not seem to work
  ⚠️ I am unable to use the latest release because of https://github.com/getsentry/sentry-electron/issues/363#issuecomment-930076320
  ⚠️ The event does not get sent to Sentry

## Versions + Platform

- [x] SDK version - `@sentry/electron@v2.4.0`
- [x] Electron version - `electron@v12.0.9`
- [x] Platform - `Linux`

## Description

We initialise Sentry like this as described [in the docs](https://docs.sentry.io/platforms/javascript/guides/electron/configuration/integrations/plugin/#offline):

```js
const { Offline: OfflineIntegration } = require('@sentry/integrations');

const { logger } = require('@sentry/utils');
logger.enable();

Sentry.init({
      dsn: '...',
      integrations: [new OfflineIntegration()],
});
```

If I turn off wifi and start the application I see the following log when I force an error from the main process:

```
Sentry Logger [Error]: Error while sending event: Error: net::ERR_INTERNET_DISCONNECTED
```

Forcing an error from the main process or from the renderer does not seem to cause an error to be logged to Sentry, even when restarting the application.

EDIT:

Forcing an error from a renderer process gives a different error, but when the machine becomes online:

```
Sentry Logger [Error]: Error while sending event: Error: net::ERR_NAME_NOT_RESOLVED
```

timfish:
Closes #369 

- Removes existing minidump queue from `BaseUploader` so there's only a single queue 
- Extends the JSON `Store` so it can parse `Date` objects 
- Adds `PersistedRequestQueue` which stores an array of queue item metadata in a `Store` and request bodies in separate files
   - Since the request body can be large and binary, we avoid serialisation 
- Adds an `ElectronOfflineNetTransport` that extends the `ElectronNetTransport` and makes it the default transport. 
   - When `sendRequest` fails, it pushes the failed request into the queue.

## Tests
Electron network emulation with `{ offline: true }` doesn't appear to work from the main process and invalid URLs take too long to timeout for reliable tests. Instead we have special DSNs which the server responds differently to.

- To test queueing, we return 429
- To test that other status drop events we return 500

Still to do:
- [x] Consider data races with `PersistedRequestQueue` with overlapping `add`/`pop`
- [x] Unit tests for `PersistedRequestQueue` since it's difficult to test all its features via e2e
- [x] Which error should result in queuing and which should be dropped? Should any errors be left uncaught or only logged?
- [x] Fix logging. I've seen "minidump sent" when it got queued

## Other
This PR also introduces caching for Electron binaries because we were getting 503 Egress exceeded errors.


src/main/transports/electron-offline-net.ts
```diff
@@ -0,0 +1,48 @@
...
+        logger.log(`Attempting to send ${request.type}`);
+        const response = await super.sendRequest(request);
+        void this.flushQueue();
+        return response;
+      } catch (e) {
+        logger.warn(e);
+      }
+    }
+
+    await this._queue.add(request);
```
AbhiPrasad, line 36: We can probably just use a simple binary semaphore to avoid contention here, maybe just have a boolean flag guarding the critical sections (add and pop)?

AbhiPrasad: Using a simple condition variable probably works just as well

timfish: Since `store.update()` is entirely sync, and `queue.add` and `queue.pop` do any update first before performing any async tasks, I don't think there are any data race issues 🤔

AbhiPrasad:
Will review this in a bit. In the mean time, https://github.com/getsentry/sentry-dotnet/issues/280 is probably useful to read through.

src/main/transports/electron-offline-net.ts
```diff
@@ -0,0 +1,58 @@
...
+import { net } from 'electron';
+import { join } from 'path';
+
+import { sentryCachePath } from '../fs';
+import { ElectronNetTransport, SentryElectronRequest } from './electron-net';
+import { PersistedRequestQueue } from './queue';
+
+/** Returns true is there's a chance we're online */
+function maybeOnline(): boolean {
+  return !('online' in net) || net.online === true;
```
AbhiPrasad, line 12: Should we also check `navigator.onLine`?

AbhiPrasad (commented):
Couldn't finish the review, but commenting what I have

src/main/transports/electron-offline-net.ts
timfish: We are in nodejs context here

timfish:
I've got some more changes incoming because I have to correctly handle different statuses from transport

`Status.Success` -> All went well 
`Status.RateLimit` -> Queue
Any other `Status` -> Drop event
`catch(Error)` -> Most probably no connection - Queue

src/main/store.ts
```diff
@@ -4,6 +4,17 @@ import { dirname, join } from 'path';
 
 import { mkdirpSync } from './fs';
 
+const dateFormat = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.*\d{0,10}Z$/;
+
+/** JSON revive function to enable de-serialization of Date objects */
+function dateReviver(_: string, value: any): any {
```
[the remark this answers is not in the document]
timfish: Tested here:
https://github.com/getsentry/sentry-electron/pull/392/files#diff-e6322789da6d7690282c4a5f1e69cc32e28469d5f77cc4e8049555063a6bf72cR40

src/main/transports/electron-offline-net.ts
AbhiPrasad: Ah yeah woooooops lol

src/main/fs.ts
```diff
@@ -24,6 +27,30 @@ export async function readFileAsync(
...
+ * @param path A relative or absolute path to the file
+ * @returns A Promise that resolves when the file has been read.
+ */
+export async function writeFileAsync(
+  path: string,
+  data: Buffer | string,
+  options?: { encoding?: string; flag?: string },
+): Promise<void> {
+  // We cannot use util.promisify here because that was only introduced in Node
+  // 8 and we need to support older Node versions.
```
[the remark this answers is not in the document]
timfish: The oldest version of Electron we now support is v2 (node v8.9.3) so all of our promise wrappers in `src/main/fs.ts` can go.

src/main/fs.ts
```diff
@@ -1,5 +1,8 @@
-import { mkdir, mkdirSync, readdir, readFile, rename, stat, Stats, statSync, unlink } from 'fs';
-import { dirname, resolve } from 'path';
+import { app } from 'electron';
+import { mkdir, mkdirSync, readdir, readFile, rename, stat, Stats, statSync, unlink, writeFile } from 'fs';
+import { dirname, join, resolve } from 'path';
+
+export const sentryCachePath = join(app.getPath('userData'), 'sentry');
```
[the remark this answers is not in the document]
timfish: Users can modify `userData` path via `app.setPath('userData', somePath)` although you'd only ever do this at the earliest possible point when the app starts and before it actually writes any Chromium data there.

It's common to change `userData` on Windows because it defaults in `~/AppData/Roaming/{{appName}}` and if this gets large, it causes issues with "roaming profiles" between machines in corporate environments. 

We've stored `sentry/scope_v2.json` in `userData` since forever and haven't had any complaints or issues.

src/main/store.ts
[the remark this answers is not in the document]
timfish: Although it might be worth having a test to ensure it works in other cases!

src/main/fs.ts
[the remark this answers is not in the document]
timfish: I'll create another PR to remove all the async wrappers

AbhiPrasad:
Will review this PR by eow, asked some other folks at Sentry to take a look to make sure we are not missing anything here.

bruno-garcia (commented):
I didn't spot this piece in the PR but are the directories where the file dumps isolated per process instance? Is there a chance two instances would pick up anothers payloads?

Seems that the URL stored enodes the auth token (and project Id?) So even if they did, they'd just be uploading the dumps from another process, is that right?

timfish:
> I didn't spot this piece in the PR but are the directories where the file dumps isolated per process instance? Is there a chance two instances would pick up anothers payloads?

Electron doesn't support multiple instances using the same `userData` directory and since we use that path to store the queue, we're freed from multi process concerns.

test/e2e/context.ts
```diff
@@ -172,6 +172,7 @@ export class TestContext {
     await this.waitForTrue(
       async () => (this.mainProcess ? !(await this.mainProcess.isRunning()) : false),
       () => 'Timeout: Waiting for app to die',
+      15_000,
```
AbhiPrasad, line 175: Is there a particular reason why we chose this number?

AbhiPrasad (approved):

test/e2e/context.ts
timfish: I changed this when the test originally tested against an unresolvable DSN. It would take forever to timeout and fail. I decided to test only against the server returning 429/500 because waiting for the connection to timeout was too flakey in the tests so this lengthy timeout can be reverted.

PR #395, opened 2021-12-03
feat: Breadcrumbs and tracing for Electron net module

### LINKED ISSUE #385: Instrument Electron `net` module
Requests are full instrumented in the browser but it would be nice to wrap the Electron [`net`](https://www.electronjs.org/docs/latest/api/net) module to have similar features in the main process too.

timfish:
Closes #385


src/common/merge.ts
```diff
@@ -1,8 +1,19 @@
...
 import { Event } from '@sentry/types';
 import * as deepMerge from 'deepmerge';
 
+/** Removes private properties fro event before merging */
+function removePrivateProperties(event: Event): void {
+  for (const span of event.spans || []) {
+    delete (span as any).spanRecorder;
+    delete span.transaction;
+  }
+}
```
timfish, line 11: I've got some questions over why this is required. 

Without it, spans contain references to the `transaction` and `spanRecorder` which include circular references and references to the `Hub` etc. This then results in stack overflow when we try and merge the event below.

AbhiPrasad: Hmm. This is a tricky problem. We need these references to make sure that `span.StartChild` works as expected. 

The `spanRecorder` reference is passed down as you add new children so we always attach spans to the correct transactions: https://github.com/getsentry/sentry-javascript/blob/cabd71348ea41748614053072481dafe44c2d669/packages/tracing/src/span.ts#L177

Same with transaction: https://github.com/getsentry/sentry-javascript/blob/cabd71348ea41748614053072481dafe44c2d669/packages/tracing/src/span.ts#L182

Can we use https://www.npmjs.com/package/deepmerge#custommerge to just do a copy instead of a recursive merge?

AbhiPrasad: Here's why `span.transaction` exists: https://github.com/getsentry/sentry-javascript/pull/2952

timfish: Ah yes that looks like a better option. 

Why is this not an issue for the other JavaScript SDKs? How do they ensure these properties are not serialised and/or sent?

src/main/integrations/net-breadcrumbs.ts
```diff
@@ -0,0 +1,189 @@
...
+            addRequestBreadcrumb('response', method, url, req, res);
+          }
+          if (tracingEnabled && span) {
+            if (res.statusCode) {
+              span.setHttpStatus(res.statusCode);
+            }
+            span.finish();
+          }
+        })
+        .once('error', function (this: ClientRequest, _error: Error): void {
```
AbhiPrasad, line 147: we don't need to `captureException` here?

src/main/integrations/net-breadcrumbs.ts
```diff
@@ -0,0 +1,189 @@
...
+   */
+  private readonly _tracing: boolean;
+
+  /**
+   * @inheritDoc
+   */
+  public constructor(options: { breadcrumbs?: boolean; tracing?: boolean } = {}) {
+    this._breadcrumbs = typeof options.breadcrumbs === 'undefined' ? true : options.breadcrumbs;
+    this._tracing = typeof options.tracing === 'undefined' ? true : options.tracing;
+  }
```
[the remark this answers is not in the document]
AbhiPrasad: We probably want to expose a couple additional options here also.

`shouldCreateSpanFromRequest` -> if a span should be created from a certain URL

`tracingOrigins` -> if sentry-trace header should be attached

src/common/merge.ts
AbhiPrasad: We do explicit merging in every other sdk, and then span serialization has an allow list: https://github.com/getsentry/sentry-javascript/blob/cabd71348ea41748614053072481dafe44c2d669/packages/tracing/src/span.ts#L316-L342

src/main/integrations/net-breadcrumbs.ts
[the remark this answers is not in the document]
timfish: > What do you think about using the `...DEFAULT_OPTIONS` pattern from the `AdditionalContext`

I thought about it but because these currently take less options (you can't pass a function here) and there are only two properties, this was much less code.  

> We probably want to expose a couple additional options here also

I'll add these. 

Once we have differing options the `Object.keys(options).reduce()` used in `AdditionalContext` becomes a little clunky and tricky to achieve in a type-safe manner. I'll give it a try though!

src/common/merge.ts
timfish: > We need these references to make sure that span.StartChild works as expected.

🤔

Since we're sending the event at this point, can we not assume the transaction is complete and safely remove `span.transaction` and `span.spanRecorder`?

AbhiPrasad: Yes we can. In fact we do this for transactions themselves:

https://github.com/getsentry/sentry-javascript/blob/cefc919560675aaedf639ad62d35c806df068537/packages/tracing/src/transaction.ts#L126-L148

timfish: So to clarify, is it safe to do keep the existing code because at this point there will be no more modification of the transaction?
```ts
delete span.spanRecorder;
delete span.transaction;
```

AbhiPrasad: Yup it is safe.

src/main/integrations/net-breadcrumbs.ts
timfish: I'm guessing we want to leave it up to the `net.request` user whether they want to capture failures as errors or not?

src/main/integrations/net-breadcrumbs.ts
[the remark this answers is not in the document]
AbhiPrasad: We need these additional options in the `http` integration as well, we'll take care of that.

src/main/integrations/net-breadcrumbs.ts
```diff
@@ -0,0 +1,204 @@
...
+  [P in keyof T]: T[P] | false;
+};
+
+type ShouldTraceFn = (method: string, url: string) => boolean;
+
+interface NetOptions {
+  breadcrumbs: boolean;
+  tracing: ShouldTraceFn;
+  tracingOrigins: ShouldTraceFn;
+}
```
AbhiPrasad, line 22: Would appreciate some comments for these options

AbhiPrasad (approved):

src/main/integrations/net-breadcrumbs.ts
timfish: Oh yes!

PR #398, opened 2021-12-06
feat: Breadcrumbs and events for child processes

### LINKED ISSUE #335: Submit events for `child-process-gone` events

timfish:
Closes #335

## Electron Normalisation
- Adds `onWebContentsCreated` and `onBrowserWindowCreated` to work around Electron v2 issue in a single place
- Add type-safe process exit reasons
- Improved `onRendererProcessGone` 
   - Allow hooking specific reasons
   - For older electron, fallback to previous limited events
   - Handle case where versions using breakpad incorrectly report `killed` for `crashed`
- Adds `onChildProcessGone` 
   - Allow hooking specific reasons
   - For older electron, fallback to previous GPU only events

## `SentryMinidump`
- Removes `addBreadcrumb` because this is now added in `ChildProcess`
- Check for and upload minidumps for crashed child processes
- Only changes the session to crashed if a minidump was found
- Add test with the GPU process minidump

## `ChildProcess`
- Add `ChildProcess` integration which is enabled by default
- Capture events for desired exit reasons (defaults to `['abnormal-exit', 'launch-failed', 'integrity-failure']`) 
- Add breadcrumbs for desired exit reasons (defaults to all)

Also fixes an issue with reporting missing count of test events


src/main/electron-normalize.ts
```diff
@@ -32,27 +32,117 @@ export function supportsFullProtocol(): boolean {
...
+export const CRASH_REASONS: ExitReason[] = ['crashed', 'oom'];
+export const ALL_REASONS: ExitReason[] = [
+  'clean-exit',
+  'abnormal-exit',
+  'killed',
+  'crashed',
+  'oom',
+  'launch-failed',
+  'integrity-failure',
+];
```
AbhiPrasad, line 53:
```suggestion
export const EXIT_REASONS = [
  'clean-exit',
  'abnormal-exit',
  'killed',
  'crashed',
  'oom',
  'launch-failed',
  'integrity-failure',
] as const;

export type ExitReason = typeof EXIT_REASONS[number];

export const CRASH_REASONS = [EXIT_REASONS[3], EXIT_REASONS[4]] as const;
```

src/main/integrations/child-process.ts
```diff
@@ -0,0 +1,113 @@
...
+      level = Severity.Warning;
+      break;
+    case 'crashed':
+    case 'oom':
+    case 'launch-failed':
+    case 'integrity-failure':
+      level = Severity.Critical;
+  }
+
+  return { message, level };
```
AbhiPrasad, line 38: this is a nit, so feel free to ignore. IMO the immutable returns are clearer to understand.

```suggestion
  switch (reason) {
    case 'abnormal-exit':
    case 'killed':
      return { message, level: Severity.Warning};
    case 'crashed':
    case 'oom':
    case 'launch-failed':
    case 'integrity-failure':
      return { message, level: Severity.Critical};
    
    return return { message, level: Severity.Debug};
  }

  return { message, level };
```

src/main/integrations/child-process.ts
```diff
@@ -0,0 +1,113 @@
+import { addBreadcrumb, captureMessage, getCurrentHub } from '@sentry/core';
+import { NodeClient } from '@sentry/node';
+import { Integration, Severity } from '@sentry/types';
+
+import { ALL_REASONS, ExitReason, onChildProcessGone, onRendererProcessGone } from '../electron-normalize';
+import { ElectronMainOptions } from '../sdk';
+
+interface ChildProcessOptions<T> {
```
AbhiPrasad, line 8: Why make this generic? It's only be used for ExitReasons right?

src/main/integrations/child-process.ts
```diff
@@ -0,0 +1,113 @@
...
+            category: 'child-process',
+            ...getMessageAndSeverity(details.reason, details.type),
+            data: details,
+          });
+        }
+      });
+
+      onRendererProcessGone(allReasons, (contents, details) => {
+        const { reason } = details;
+        const getName = (): string => options?.getRendererName?.(contents) || 'renderer';
```
AbhiPrasad, line 93: Why is `getName` a function?

AbhiPrasad (approved):
cool that we can track the gpu stuff!

src/main/integrations/child-process.ts
timfish: No I agree here!

src/main/integrations/child-process.ts
timfish: Yeah the name will be required at this point, so no reason to lazily evaluate it

src/main/electron-normalize.ts
timfish: Not sure I agree with this one. 

The array indexing makes it easier to break in the future and it's not immediately clear which events are crashes. 

They both transpile down to a similar length JavaScript.

src/main/integrations/child-process.ts
timfish: watch me try and attempt that across PRs 🤣

src/main/electron-normalize.ts
AbhiPrasad: yeah I was iffy about the clarity from `CRASH_REASONS` as well. I think the `type ExitReason` is good though.

we can just make it

```ts
export const CRASH_REASONS: Readonly<ExitReason[]> = ['crashed', 'oom'] as const;
```

src/main/integrations/child-process.ts
AbhiPrasad: lol

We can just do it in another PR, np. Step by step!

src/types.ts
```diff
@@ -0,0 +1,9 @@
+type Optional<T, K extends keyof T> = Pick<Partial<T>, K> & Omit<T, K>;
+
+type OrBool<T> = {
```
AbhiPrasad, line 3: These don't have to be exported?

AbhiPrasad (approved):

src/types.ts
timfish: Apparently not!

AbhiPrasad: 😮

PR #402, opened 2021-12-20
Fix: Wait before counting files in unit tests

timfish:
The units tests do some checks to ensure that files are correctly cleaned up. However, this cleanup occurs asynchronously so we should delay to ensure this has had enough time to complete.

### DIFF
--- a/test/unit/persist-queue.test.ts
+++ b/test/unit/persist-queue.test.ts
@@ -9,7 +9,16 @@ import { walkSync } from '../e2e/utils';
 should();
 use(chaiAsPromised);
 
-function expectFilesInDirectory(dir: string, count: number): void {
+function delay(ms: number): Promise<void> {
+  return new Promise((resolve) => {
+    setTimeout(resolve, ms);
+  });
+}
+
+async function expectFilesInDirectory(dir: string, count: number): Promise<void> {
+  // We delay because store flushing is async and not waited on
+  await delay(500);
+
   const found = Array.from(walkSync(dir)).length;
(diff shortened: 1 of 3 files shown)


test/unit/persist-queue.test.ts
```diff
@@ -9,7 +9,16 @@ import { walkSync } from '../e2e/utils';
 should();
 use(chaiAsPromised);
 
-function expectFilesInDirectory(dir: string, count: number): void {
+function delay(ms: number): Promise<void> {
+  return new Promise((resolve) => {
```
AbhiPrasad, line 13: Nit: should we preemptively put this in a `helpers.ts`?

AbhiPrasad (approved):
Good old timeouts in tests

test/unit/persist-queue.test.ts
timfish: ah, almost certainly. I think we already have `delay` elsewhere.

AbhiPrasad (approved):
:shipit:

timfish:
There are still some flakey e2e failures on macOS. It looks like the main process has native crashed whereas we're expecting a renderer crash.

I don't have my Mac with me until the new year so I'm unable to dig any deeper locally!

AbhiPrasad:
I'm also out on vacations for the rest of the year - but if you don't get your hands on a macos machine, I can help debug when I get back in the new year!

Happy holidays @timfish!

timfish:
I ran the tests again and they passed. Just massive flake on the macOS tests.

Issue #403, opened 2021-12-29
Deepmerge import is broken in vite/rollup


artplan1:
- [x] Review the documentation: https://docs.sentry.io/platforms/javascript/electron/
- [x] Search for existing issues: https://github.com/getsentry/sentry-electron/issues
- [x] Use the latest release: https://github.com/getsentry/sentry-electron/releases
- [ ] Provide a link to the affected event from your Sentry account

## Versions + Platform

- [x] SDK version - `@sentry/electron@v3.0.0-beta-3`
- [x] Electron version - `electron@v16.0.5`
- [x] Platform - `Windows/macOS/Linux`
- [x] Bundler - `Vite/Rollup`

## Description

Errors:
```bash
Cannot call a namespace ('deepMerge')
```

```bash
Sentry Logger [Error]: SentryError2: Event processing pipeline threw an error, original event will not be sent. Details have been sent as a new event.
Reason: TypeError: deepMerge is not a function
```

The way deepmerge is imported at https://github.com/getsentry/sentry-electron/blob/12058e46133f762bb7bdc873982727a3da813e88/src/common/merge.ts#L3 works incorrectly in Vite/Rollup - https://github.com/TehShrike/deepmerge/issues/102

Similar issue: https://github.com/vitejs/vite/issues/2261#issuecomment-808961321

Fix is to change import to intended: https://github.com/TehShrike/deepmerge/issues/102#issuecomment-384975231
```js
import deepmerge from 'deepmerge'
```

timfish:
I've added an example test for vite/rollup and it works without any SDK changes:
https://github.com/getsentry/sentry-electron/pull/406

If you can supply an example where it doesn't work I'd be happy to take a look!

PR #404, opened 2021-12-29
Fixes deepmerge import

artplan1:
Fixes https://github.com/getsentry/sentry-electron/issues/403


timfish:
Thanks for looking into this. There are however a couple of issues:

- This change breaks the TypeScript build
- Ideally we should add an integration test to ensure this doesn't get broken again in the future

Do you have a minimal reproduction of this issue that could be the basis of a test?

timfish:
Closing in favour of #406

PR #417, opened 2022-02-18
Fix: macos test failures

timfish:
This PR:
- Fixes a double quote typo in child process breadcrumbs + relevant tests
- Fixes e2e app directory deleting on macOS which was allowing context leakage between test runs
- Extends some `quit` timeouts which caused intermittent failures on macOS
- Adds a reusable means to track renderer properties so they're accessible after the window is destroyed
  - Fixed an issue observed on macOS where occasionally `webContents.getURL()` would throw due to destroyed window
  - Updated `ElectronBreadcrumbs` integration to use stored window state so `data` is now included even for `destroyed` events
- Reinstates macOS CI tests


src/main/renderers.ts
```diff
@@ -0,0 +1,58 @@
...
+    contents.on('did-navigate', (_, url) => updateUrl(id, url));
+    contents.on('did-navigate-in-page', (_, url) => updateUrl(id, url));
+    contents.on('page-title-updated', (_, title) => updateTitle(id, title));
+
+    contents.on('destroyed', () => {
+      // We need to delay since consumers of this API sometimes need to
+      // access the state shortly after a renderer is destroyed
+      setTimeout(() => {
+        renderers.delete(id);
+      }, 5000);
```
AbhiPrasad, line 50: it has to be timeout for sure? We can't wait a few event loop ticks?

src/main/renderers.ts
```diff
@@ -0,0 +1,58 @@
...
+
+let RENDERERS: Map<number, Renderer> | undefined;
+
+/**
+ * Enable tracking of renderer properties via events
+ *
+ * This allows us to get the last known state of a renderer even if it's been destroyed
+ **/
+export function trackRendererProperties(): void {
+  if (RENDERERS) {
```
AbhiPrasad, line 20: this is clever to use `RENDERERS` as a flag to make sure we only add the event emitters once, but I wonder how much it reduces readability.

Wonder if an explicit boolean flag + defining `const RENDERERS = new Map<number, Renderer>();` with would help with that. I'm torn, so feel free to ignore this.

I guess this way though the map creation is lazy though, which is another added plus.

AbhiPrasad (approved):
nice, extracting the renderer info into shared state is a great solution for this!

src/main/renderers.ts
timfish: Yes, so `trackRendererProperties` is called from two different integrations that require this data and wont be called at all if you don't use those integrations..

timfish:
I've included an update to the changelog since I plan on releasing this as 3.0.1

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 3.0.1
+
+- fix: Fixes a potential issue on macOS where the window URL is not accessible after crash
```
AbhiPrasad, line 7:
```suggestion
- fix: Fixes a potential issue on macOS where the window URL is not accessible after crash (#417)
```

AbhiPrasad (approved):

PR #419, opened 2022-02-23
fix: Fix broken serialization of node transaction spans

### LINKED ISSUE #418: Custom Transactions are not showing in the Performance tab
## Versions + Platform

- [ ] SDK version - `@sentry/electron@v3.0.0`
- [ ] Electron version - `electron@v13.6.3`
- [ ] Platform - `macOS`

## Description

`

    "@sentry/electron": "^3.0.0",
    "@sentry/integrations": "^6.17.0",
    "@sentry/node": "^6.17.0",
    "@sentry/react": "^6.17.0",
    "@sentry/tracing": "^6.17.0",
    "@sentry/types": "^6.17.0",
`

I am instrumenting custom transactions based on the docs however the transaction does not appear in the performance dashboard.
It is is capturing the default `Request` transaction from electron but not my custom one.
I do see in the logs that the transaction is started & finished but never sent to sentry.

My sampleRate & traceSampleRate is set to 1.

`

      enabled: true,
      debug: true,
      environment: Environment.Development,
      sampleRate: 1,
      tracesSampleRate: 1,
`

my code looks like:

`

    const Sentry = require('@sentry/node');

    async initSequence() {

      const transaction = Sentry.startTransaction({ name: 'InitSequence', op: 'task' });
      Sentry.getCurrentHub().configureScope((scope) => scope.setSpan(transaction));
    
      const initializeServicesSpan = transaction.startChild({ op: 'initializeServices' });
      await this.initializeServices();
      initializeServicesSpan.finish();
      transaction.finish();
    }
`

Here is what I see in logs
  
 Sentry Logger [Log]: [Tracing] starting task transaction - InitSequence 
 Sentry Logger [Log]: [Tracing] Finishing task transaction: InitSequence. 
 Sentry Logger [Log]: Sending 'transaction' request
 Sentry Logger [Log]: Successfully sent

timfish:
Fixes #418

The Electron SDK has a `mergeEvents` function which uses `deepmerge`. This saves having a large manual merge function which has to be kept up to date if any of the event objects change. 

The Sentry tracing `Span` has a custom `toJSON` implementation to convert camelCase properties to snake_case to match the Sentry API guidelines. When the event gets merged with the defaults this function is lost.

This PR:
- Manually copies the spans so their `toJSON` function is still there
- Adds an e2e test to ensure this doesn't break in the future
- Improves the e2e event normalisation code so it doesn't add properties

### DIFF
--- a/test/e2e/test-apps/other/custom-tracing/src/main.js
+++ b/test/e2e/test-apps/other/custom-tracing/src/main.js
@@ -2,8 +2,7 @@
 const { app } = require('electron');
 
 const Sentry = require('@sentry/electron');
-// eslint-disable-next-line no-unused-vars
-const Tracing = require('@sentry/tracing');
+require('@sentry/tracing');
 
 Sentry.init({
   dsn: '__DSN__',
@@ -15,8 +14,6 @@ Sentry.init({
 
 async function doTransaction() {
   const transaction = Sentry.startTransaction({ name: 'InitSequence', op: 'task' });
-  Sentry.configureScope((scope) => scope.setSpan(transaction));
-
   const initializeServicesSpan = transaction.startChild({ op: 'initializeServices' });
 
   setTimeout(() => {
--- a/test/e2e/test-apps/other/custom-tracing/src/main.js
+++ b/test/e2e/test-apps/other/custom-tracing/src/main.js
@@ -1,0 +1,28 @@
+// eslint-disable-next-line import/order
+const { app } = require('electron');
+
+const Sentry = require('@sentry/electron');
+// eslint-disable-next-line no-unused-vars
+const Tracing = require('@sentry/tracing');
+
+Sentry.init({
+  dsn: '__DSN__',
+  debug: true,
+  autoSessionTracking: false,
+  tracesSampleRate: 1.0,
+  onFatalError: () => {},
+});
+
+async function doTransaction() {
+  const transaction = Sentry.startTransaction({ name: 'InitSequence', op: 'task' });
+  Sentry.configureScope((scope) => scope.setSpan(transaction));
+
+  const initializeServicesSpan = transaction.startChild({ op: 'initializeServices' });
+
+  setTimeout(() => {
+    initializeServicesSpan.finish();
+    transaction.finish();
+  }, 500);
+}
+
+app.on('ready', () => doTransaction());
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 3.0.2
+
+- fix: Fix broken serialization of node transaction spans (#419)
+
 ## 3.0.1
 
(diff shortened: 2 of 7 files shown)


test/e2e/test-apps/other/custom-tracing/src/main.js
```diff
@@ -0,0 +1,28 @@
+// eslint-disable-next-line import/order
+const { app } = require('electron');
+
+const Sentry = require('@sentry/electron');
+// eslint-disable-next-line no-unused-vars
+const Tracing = require('@sentry/tracing');
```
AbhiPrasad, line 6: A simple require for the side effect should be enough?

```suggestion
require('@sentry/tracing');
```

test/e2e/test-apps/other/custom-tracing/src/main.js
```diff
@@ -0,0 +1,28 @@
...
+  dsn: '__DSN__',
+  debug: true,
+  autoSessionTracking: false,
+  tracesSampleRate: 1.0,
+  onFatalError: () => {},
+});
+
+async function doTransaction() {
+  const transaction = Sentry.startTransaction({ name: 'InitSequence', op: 'task' });
+  Sentry.configureScope((scope) => scope.setSpan(transaction));
```
AbhiPrasad, line 18: Why do we need to set the transaction on the scope?

AbhiPrasad (approved):
Do we need to do similar for sessions as well? I assume it's not a problem for now because sessions don't got through the event processors or from renderer -> main process.

test/e2e/test-apps/other/custom-tracing/src/main.js
timfish: I've just tested it and it works without `@sentry/tracing` entirely 🤔

test/e2e/test-apps/other/custom-tracing/src/main.js
timfish: Ah, the docs say to do this but now I've read them again I see this is only important for linking them to errors!

test/e2e/test-apps/other/custom-tracing/src/main.js
AbhiPrasad: 🤔 Are we doing the require call somewhere else?

timfish: No!  🤣

Could it be this?
https://github.com/getsentry/sentry-javascript/blob/6809dd2ae41f81e3b5d3e3c8076f2c6bcb9cfbfb/packages/node/src/handlers.ts#L4

timfish:
> Do we need to do similar for sessions as well? 

I think not. Sessions don't go through the event processors and we don't try and do anything with them at all.

test/e2e/test-apps/other/custom-tracing/src/main.js
AbhiPrasad: It’s the first one! Lol

PR #428, opened 2022-03-05
fix: Lazily initialise IPC in renderer

### LINKED ISSUE #427: Importing @sentry/electron without calling init() yields an error

## Versions + Platform

- [x] SDK version - `@sentry/electron@3.0.3`
- [x] Electron version - `electron@17.1.0`
- [x] Platform - `Linux`

## Description

We call `init()` conditionally, so that our users have an option to turn off error reporting. This worked fine in `@sentry/electron@2.5.4`, but in `@sentry/electron@3.0.3`, when `init()` is not called, we get this console error in the renderer process from #377:

```
Fetch API cannot load sentry-ipc://sentry-electron.ping. URL scheme "sentry-ipc" is not supported.", source: …/node_modules/@sentry/electron/renderer/ipc.js (15)
Sentry SDK failed to establish connection with the Electron main process.
 - Ensure you have initialized the SDK in the main process
 - If your renderers use custom sessions, be sure to set 'getSessions' in the main process options
 - If you are bundling your main process code and using Electron < v5, you'll need to manually configure a preload script", source: …/node_modules/@sentry/electron/renderer/ipc.js (15)
```

`@sentry/electron` do nothing before `init()` is called (like the documentation [says](https://github.com/getsentry/sentry-electron/#usage)).

### Minimal reproducible example

**`index.js`**

```js
const Sentry = require("@sentry/electron");
const { app, BrowserWindow } = require("electron");

if (false /* we decided not to */) {
  Sentry.init({ dsn: "__DSN__" });
}

app.whenReady().then(() => {
  const win = new BrowserWindow({
    webPreferences: {
      preload: require.resolve("./preload.js"),
    },
  });
  win.loadURL("about:blank");
});
```

**`preload.js`**

```js
const Sentry = require("@sentry/electron");

if (false /* we decided not to */) {
  Sentry.init({ dsn: "__DSN__" });
}
```

**`package.json`**

```json
{
  "dependencies": {
    "@sentry/electron": "^3.0.3",
    "electron": "^17.1.0"
  }
}

### PR DESCRIPTION
**Pull request by timfish**
Fixes #427

### DIFF
--- a/src/renderer/ipc.ts
+++ b/src/renderer/ipc.ts
@@ -2,12 +2,10 @@
 /* eslint-disable no-console */
 import { logger } from '@sentry/utils';
 
-import { ensureProcess, IPCChannel, IPCInterface, PROTOCOL_SCHEME } from '../common';
+import { IPCChannel, IPCInterface, PROTOCOL_SCHEME } from '../common';
 
 /** Gets the available IPC implementation */
 function getImplementation(): IPCInterface {
-  ensureProcess('renderer');
-
   // Favour IPC if it's been exposed by a preload script
   if (window.__SENTRY_IPC__) {
     return window.__SENTRY_IPC__;
@@ -38,10 +36,18 @@ function getImplementation(): IPCInterface {
   };
 }
 
+let cachedInterface: IPCInterface | undefined;
+
 /**
  * Renderer IPC interface
  *
  * Favours IPC if its been exposed via a preload script but will
  * fallback to custom protocol and fetch is IPC is not available
  */
-export const IPC: IPCInterface = getImplementation();
+export function getIPC(): IPCInterface {
+  if (!cachedInterface) {
+    cachedInterface = getImplementation();
+  }
+
+  return cachedInterface;
+}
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/renderer/ipc.ts [hunk 1]
```diff
@@ -38,10 +36,18 @@ function getImplementation(): IPCInterface {
...
 /**
  * Renderer IPC interface
  *
  * Favours IPC if its been exposed via a preload script but will
  * fallback to custom protocol and fetch is IPC is not available
  */
-export const IPC: IPCInterface = getImplementation();
+export function getIPC(): IPCInterface {
+  if (!cachedInterface) {
+    cachedInterface = getImplementation();
```
AbhiPrasad, line 49: Will there be any worries about the cache getting stale? Like some condition where `window.__SENTRY_IPC__` somehow changes?

timfish: We were only ever calling `getImplementation()` once before. This change just does it lazily rather than at parse time.

AbhiPrasad (approved):

PR #430, opened 2022-03-06
Fix: Bug when listing latest Electron versions

timfish:
A bug meant not all versions were getting listed correctly.

https://github.com/getsentry/sentry-electron/pull/429/files


AbhiPrasad (approved):
ref: https://github.com/timfish/electron-latest-versions/commit/89072089d58a29fd36ef913f56c446a4b6335450

PR #432, opened 2022-03-06
fix: Use `esModuleInterop` for `deepmerge`

### LINKED ISSUE #431: Cannot find module "deepmerge"
Hey there, I'm still seeing this `Cannot find module "deepmerge"` error in the compiled version of my app.
I can't load sentry at all and can't find any related issues beyond the PRs that attempted to fix the original problem.

This problem only applies after building the app using electron-builder.

I've confirmed that the issue lies at the original spot:
https://github.com/getsentry/sentry-electron/issues/403
https://github.com/getsentry/sentry-electron/pull/406

`electron/esm/common/merge.js`
I was able to successfully start my application with Sentry if the deepMerge import was changed to:

```
// const deepMerge = require('deepmerge');
import deepMerge from 'deepmerge';
```

The `merge.ts` source code has `const deepMerge = require('deepmerge');`; but this is incompatible with the esm flavor.

## Versions + Platform

```
    "electron": "17.1.0",
    "electron-builder": "22.14.13",
    "vite": "^2.8.6",
    "@sentry/electron": "^3.0.3",
```

This impacts both Mac and Window builds.

Seems like this is the problem:
https://2ality.com/2019/10/hybrid-npm-packages.html
https://redfin.engineering/node-modules-at-war-why-commonjs-and-es-modules-cant-get-along-9617135eeca1

timfish:
Fixes #431

It looks like we need to use the TypeScript `esModuleInterop` compiler option for `deepmerge`

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,6 +9,5 @@
   ],
   "compilerOptions": {
     "rootDir": ".",
-    "esModuleInterop": false,
   }
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,8 +1,14 @@
 {
   "extends": "./tsconfig.build.json",
-  "include": ["src/**/*.ts", "test/**/*.ts"],
-  "exclude": ["dist"],
+  "include": [
+    "src/**/*.ts",
+    "test/**/*.ts"
+  ],
+  "exclude": [
+    "dist"
+  ],
   "compilerOptions": {
-    "rootDir": "."
+    "rootDir": ".",
+    "esModuleInterop": false,
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+## 3.0.4
+
+- fix: Use esModuleInterop for deepmerge (#432)
+- fix: Lazily initialise IPC in renderer (#428)
+
 ## 3.0.3
 
 - fix: Don't add empty breadcrumbs (#425)
--- a/src/common/merge.ts
+++ b/src/common/merge.ts
@@ -1,6 +1,5 @@
 import { Event } from '@sentry/types';
-// import * as deepmerge does not work with ES modules
-const deepMerge = require('deepmerge');
+import deepMerge from 'deepmerge';
 
 /** Removes private properties from event before merging */
 function removePrivateProperties(event: Event): void {
--- a/test/e2e/index.test.ts
+++ b/test/e2e/index.test.ts
@@ -1,6 +1,6 @@
 import { should, use } from 'chai';
-import * as chaiAsPromised from 'chai-as-promised';
-import * as chaiSubset from 'chai-subset';
+import chaiAsPromised from 'chai-as-promised';
+import chaiSubset from 'chai-subset';
 import { join } from 'path';
 
 import { TestContext } from './context';
--- a/test/e2e/recipe/parser.ts
+++ b/test/e2e/recipe/parser.ts
@@ -1,7 +1,7 @@
 import { Event, Session } from '@sentry/types';
 import { readFileSync } from 'fs';
 import { dirname, sep } from 'path';
-import * as YAML from 'yaml';
+import YAML from 'yaml';
 
 import { TestServerEvent } from '../server';
 import { walkSync } from '../utils';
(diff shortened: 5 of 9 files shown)


AbhiPrasad (commented):
Requesting a review from Katie cause shes knows about this a lot more than I do

timfish:
An alternative would be to vendor `deepmerge` and convert it to TypeScript?

AbhiPrasad:
> An alternative would be to vendor `deepmerge` and convert it to TypeScript?

Yeah we've been thinking about doing the same with `localforage`: https://github.com/getsentry/sentry-javascript/blob/845aada785a956e6348d87da9ff183aec24b5add/packages/integrations/src/offline.ts#L5

timfish:
> Yeah we've been thinking about doing the same with `localforage`

Since `indexedDb` is supported on everything apart from IE, it might be worth just migrating to something `indexedDb` based rather than a library that abstracts over `localStorage/indexedDb`.

I've used [`idb`](https://www.npmjs.com/package/idb) quite a bit and can recommend it.

lobsterkatie:
> > Yeah we've been thinking about doing the same with `localforage`
> 
> Since `indexedDb` is supported on everything apart from IE, it might be worth just migrating to something `indexedDb` based rather than a library that abstracts over `localStorage/indexedDb`.
> 
> I've used [`idb`](https://www.npmjs.com/package/idb) quite a bit and can recommend it.

Yeah, we were musing the other day that the offline integration could probably just be a transport, based on native APIs, allowing us to finaly shed our `localforage` dependency, so it can't cause us any more grief.

tsconfig.json
```diff
@@ -1,8 +1,14 @@
...
+    "src/**/*.ts",
+    "test/**/*.ts"
+  ],
+  "exclude": [
+    "dist"
+  ],
   "compilerOptions": {
-    "rootDir": "."
+    "rootDir": ".",
+    "esModuleInterop": false,
```
lobsterkatie, line 12: Why turn this back off here?

lobsterkatie (commented):
@timfish, could you please include in the PR description what problem this is fixing? What's the catalyst for changing this now?

P.S., I can see that these configs are based on our old tsconfigs, which were kind of a mess. I cleaned it up and streamlined them quite a bit in https://github.com/getsentry/sentry-javascript/pull/4178, if you're interested.

tsconfig.json
jasonlor: Not the original author, but I think this is to fix the issue I filed: https://github.com/getsentry/sentry-electron/issues/431

I can't add sentry to my compiled application (that uses ES modules) due to `Cannot find module "deepmerge"` issue in the compiled version of `/esm/common/merge.js`

timfish:
> include in the PR description

Somehow missed that!

> I cleaned it up and streamlined them quite a bit... if you're interested

Oooh yeah, I'll take a look.

jasonlor:
> > include in the PR description
> 
> Somehow missed that!
> 
> > I cleaned it up and streamlined them quite a bit... if you're interested
> 
> Oooh yeah, I'll take a look.

@timfish Not sure if I configured this accurately but I updated the package.json to:
`"@sentry/electron": "https://github.com/timfish/sentry#try-esModuleInterop",`

I checked `node_modules` and sentry doesn't appear to be compiled when I run `yarn compile`.
Vite returns this error:

```
vite v2.8.6 building for production...
✓ 1 modules transformed.
[vite]: Rollup failed to resolve import "@sentry/electron/main" from "src/index.ts".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
error during build:
Error: [vite]: Rollup failed to resolve import "@sentry/electron/main" from "src/index.ts".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
```

timfish:
@jasonlor we don't check the build output into source control so it won't work as a git dependency.

It does get built in Github Actions though, so you can download the output artifact [from here](https://github.com/getsentry/sentry-electron/actions/runs/1942487751) from a successful run and add as a dependency like:

```json
  "@sentry/electron": "../../path/to/sentry-electron-v3.0.3.tgz",
```

jasonlor:
@timfish I can confirm that this resolves the issue!

AbhiPrasad (approved):

PR #435, opened 2022-03-10
fix: getSessions should be typed as optional

MarshallOfSound:
Documented as optional but typed as required

### DIFF
--- a/src/main/sdk.ts
+++ b/src/main/sdk.ts
@@ -49,7 +49,7 @@ export interface ElectronMainOptions extends NodeOptions {
    *
    * Defaults to () => [session.defaultSession]
    */
-  getSessions: () => Session[];
+  getSessions?: () => Session[];
   /**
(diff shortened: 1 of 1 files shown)


timfish:
This is already optional via the parameters being `Partial<ElectronMainOptions>`.

https://github.com/getsentry/sentry-electron/blob/b11442f04012c97fb3e05ffa73dce5916e73c3d0/src/main/sdk.ts#L70

Any reason to make the property optional too?

MarshallOfSound:
When `ElectronMainOptions` is re-used as a type in a helper function E>g. `getOptions(): ElectronMainOptions` it is vaguely unexpected behavior that an options struct wouldn't be used directly and would instead be `Partial<T>` in a parameter interface.  Can we instead just type `ElectronMainOptions` correctly so that the `Partial` isn't needed?

timfish:
Understood! Closing in favour of #448

PR #436, opened 2022-03-10
fix: ElectronNetTransport stripping query params from requests

CliftonH:
Ensuring that the query params are present after being stripped out by `_getRequestOptions` [here](https://github.com/getsentry/sentry-javascript/blob/2d3b86122a618500b0c7a1d1017e1d561f789c1f/packages/node/src/transports/base/index.ts#L121).


timfish:
I heard that you have worked around this elsewhere so I'll close this for now.

Get back to me if this isn't the case!

PR #441, opened 2022-03-21
Fix exponential backoff delay overflowing the maximum delay in setTimeout, causing a high number of requests and errors

leolabs:
The `setTimeout` function has an interesting implementation detail where, if the delay is bigger than the biggest signed 32-bit integer, it will throw an error and run the callback after 1ms.

The error looks something like this:

```
TimeoutOverflowWarning: Infinity does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
```

Since this error is also reported to Sentry by default, this causes a huge number of errors which can block the main thread.

My workaround for this is to limit the delay to 2,000,000,000 ms so it never overflows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
-## Unreleased
+## 3.0.5
+
+- fix: Limit retryDelay to avoid integer overflows in setTimeout (#441)
 
 ## 3.0.4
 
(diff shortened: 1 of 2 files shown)


timfish:
Thanks for the PR.

Can you fix the linting error and add a line to `CHANGELOG.md` under the  `3.0.5` heading? 

I'll include this fix in the release going out today!

timfish (approved):

timfish:
Thanks again for the PR. 

Subscribe to [this issue](https://github.com/getsentry/publish/issues/939) to get notified of the release.

Issue #450, opened 2022-03-29
app.exit() called even if preventDefault() used in will-quit event


phallguy:
- [X] Review the documentation: https://docs.sentry.io/platforms/javascript/electron/
- [X] Search for existing issues: https://github.com/getsentry/sentry-electron/issues
- [X] Use the latest release: https://github.com/getsentry/sentry-electron/releases
- [ ] Provide a link to the affected event from your Sentry account

## Versions + Platform

- [X] SDK version - `@sentry/electron@v3.0.5`
- [X] Electron version - `electron@v16.0.10`
- [X] Platform - All `Windows/macOS/Linux`

## Description

We listen for the `will-quit` event to perform some cleanup before the app is shut down. In particular flushing data and waiting for pending tasks to complete. This may take a few minutes so we use `preventDefault()` to stop the process and perform our cleanup, then exit gracefully.

[MainProcessEssion _exitHandler](https://github.com/getsentry/sentry-electron/blob/e7b9505f103cbf48a5def1a3002e515d6195154a/src/main/integrations/main-process-session.ts#L55) takes a similar approach. However it cleans up much faster than our cleanup code and calls `app.exit()` before we've completed our cleanup work causing the app to shutdown immediately. 

I've confirmed that adding a this to the exit handler allows it to exit gracefully while still allowing other handlers to block stop the app from quitting.

```
if(event.defaultPrevented) {
   return;
}
```

timfish:
This has been released in 3.0.7. Sorry for the delay!

PR #451, opened 2022-03-29
Don't exit if preventDefault used in will-quit event

phallguy:
Fixes https://github.com/getsentry/sentry-electron/issues/450


timfish:
Ah yes, we force the session handling to be the last `will-quit` handler but we're not checking if any of the other stop the exit.

Thanks for the PR!

phallguy:
Sure thing...I didn't check in our quit handler either before forcing it to quit when we finished our work. I mean...who does that? 😄

timfish (approved):

PR #479, opened 2022-05-04
chore: Update logo for dark or light theme

mattjohnsonpint:
<s>Update logo to use media queries so it looks good on both dark and light themes.

See https://github.com/getsentry/sentry/pull/34229 for screenshot of effect, or view readme on this branch after [changing your theme](https://github.com/settings/appearance).
</s>

See comments below.  Decided to use purple Sentry logo instead, so it renders well both on GitHub and NPM.

#skip-changelog


timfish:
What does this look like when the package is published to npm?

I've messed around with this before but realised that you can end up with the wrong image on npm/pypi etc.

timfish:
Confirmed by pasting the HTML via dev tools with my OS default set to dark:

image

mattjohnsonpint:
Oh. Let me see what can be done about that. I already merged this on a ton of other repos, so hopefully won't have to unwind. 🙂

timfish:
It's tricky and the results vary by package directory. For example, it seems Nuget doesn't display HTML from the readme any way!

mattjohnsonpint:
Looks like there is some built-in support from Github's markdown:

https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#specifying-the-theme-an-image-is-shown-to

I'm going to try that out and will be back shortly with an update.

mattjohnsonpint:
From testing and research, the built-in support apparently has major problems, and doesn't work when the image is wrapped in a hyperlink.  NPM would show both images anyway.

I'm going to leave the default implementation for most of Sentry's repositories, and make adjustments where the repo is used with a package registry.  Where media queries aren't supported (like NPM), I'll use Sentry's purple logo (which is better on black background than the black logo, but still not quite as clear as white).  Where HTML isn't supported (like Nuget), I'll revert to using plain markdown - which doesn't support centering, but keeps it rendering clean.

@timfish - Thanks for keeping me on my toes! 👍

timfish (approved):

timfish:
> Thanks for keeping me on my toes!

Yes, sorry. I responded here and then responded in the Rust PR and then noticed all the other PRs too!

PR #471, opened 2022-04-20
feat: Update to v7 JavaScript SDKs 

timfish:
Upgrading to v7 of the Sentry JavaScript SDKs:
- Migrate enums to strings
- Migrate away from `forget` helper
- Adjust exports for `isolatedModules`
- Rewrite transports to use new style API
- Refactor minidump submission since the underlying SDKs now support attachments!

Closes #496, #481, #487, #489 and #497

PR #484, opened 2022-05-11
fix: Pass correct event to beforeSend

### LINKED ISSUE #481: SentryMinidump: native crash report does not contain expected scope / context if a beforeSend function is supplied

## Versions + Platform

- [x] Electron version - `electron@v16.0.10`
- [x] Platform - `Windows`

## Description

As in title - if you supply a `beforeSend` function then the default SentryMinidump integration will not correctly fill in the scope on the event when sending a crash report.

This may be intentional, but I couldn't find docs stating this. The problematic line is this one - https://github.com/getsentry/sentry-electron/blob/master/src/main/integrations/sentry-minidump/index.ts#L234 

which instead of giving the `beforeSend` function the `newEvent` that has been filled with scope, it passes the original `event`.

Fixing should be trivially to pass `newEvent` to `beforeSend` (we have patched this in locally and it seems to be working in the way we expect).

timfish:
Fixes #481

### DIFF
--- a/src/main/integrations/sentry-minidump/index.ts
+++ b/src/main/integrations/sentry-minidump/index.ts
@@ -231,7 +231,7 @@ export class SentryMinidump implements Integration {
           }
 
           if (newEvent && beforeSend) {
-            const beforeSendResult = await this._ensureBeforeSendRv(beforeSend(event));
+            const beforeSendResult = await this._ensureBeforeSendRv(beforeSend(newEvent));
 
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):

timfish:
I would normally add a test for this but once #471 is merged in, we'll be going through `beforeSend` and `sampleRate` in `@sentry/core` so much of this code can go.

Issue #507, opened 2022-07-08
Errors while process crashes in native add-on


haoxi911:
## Versions + Platform

- [x]  SDK version - @sentry/electron@v3.0.7
- [x]  Electron version - electron@v19.0.8
- [x]  Platform - macOS

## Description

Can anyone help to take a look at this [issue](https://sentry.io/organizations/simplified-it-products/issues/3411630838/?project=6468531&query=is%3Aunresolved)? It says there is an error while processing the event: **sdkProcessingMetadata: Discarded unknown attribute**

When I run `npx sentry-cli sourcemaps explain b280d03d3e6c486691baa0d38b602f6d`, it gives the following output:
```
✔ Fetched data for event: b280d03d3e6c486691baa0d38b602f6d
✔ Event has release name: darwin@4.5.5
✔ Event has a valid exception present
✔ Event has a valid stacktrace present
✖ Event exception stacktrace top frame has incorrect abs_path (valid url is required). Found /Users/Kevin/Projects/Electron/.../addon/src/feature/mac/PhotosSearch.m
```

The symbol was resolved properly though, just curious if we can get rid of these errors.

AbhiPrasad:
We have to make sure `sdkProcessingMetadata` is stripped before it is sent to the server.

Perhaps we have to strip it here? https://github.com/getsentry/sentry-electron/blob/f11ab00774db8e85417d08f5f243bf9c290520a0/src/main/integrations/electron-minidump.ts#L162

PR #508, opened 2022-07-08
fix: Delete sdkProcessingMetadata in minidump pipeline

AbhiPrasad:
Fixes https://github.com/getsentry/sentry-electron/issues/507

See: https://github.com/getsentry/sentry-javascript/blob/f372d842ba39d5f107c055e97c5219a05c543dba/packages/core/src/envelope.ts?q=sdkProcessingMetadata#L86

Delete `event.sdkProcessingMetadata` before the event is sent as a minidump to Sentry.

### DIFF
--- a/src/main/integrations/electron-minidump.ts
+++ b/src/main/integrations/electron-minidump.ts
@@ -169,6 +169,8 @@ export class ElectronMinidump implements Integration {
     // Apply the scope to the event
     await scope.applyToEvent(event);
 
+    delete event.sdkProcessingMetadata;
+
     // Normalise paths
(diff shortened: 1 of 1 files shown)


timfish (approved):

timfish:
Isn't `sdkProcessingMetadata` gone in v7?

AbhiPrasad:
It was... until it wasn't. We use it for some general metadata that we need for the Dynamic (server-side) sampling project we are working on. See more details here: https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/

haoxi911:
I don't understand what that for, but thank you for quick fix and merge!

PR #509, opened 2022-07-09
fix: Correctly parse mixed chrome/node renderer stack traces

### LINKED ISSUE #308: Missing stacktrace in renderer process when using Node.js API integration

## Versions + Platform

- [x] SDK version - `@sentry/electron@v2.4.0`
- [x] Electron version - `electron@v11.0.4`
- [x] Platform - `macOS`

## Description

When capturing a exception thrown by the Node integration in the renderer process, the reported stacktrace is empty.

Sample code to reproduce:
```ts
  const exec = promisify(childProcess.exec);
  const command = `commandThatFails`;
  let stdout: string;
  try {
    const response = await exec(command);
    stdout = response.stdout;
  } catch (error) {
    console.log('got error', error);
    stdout = error.stdout;
    if (!stdout) {
      Sentry.captureException(error);
    }
  }
```

The console prints the following (we can see here the value of error.stack):
```
got error Error: Command failed: commandThatFails
/bin/sh: commandThatFails: command not found

    at ChildProcess.exithandler (child_process.js:312)
    at ChildProcess.emit (events.js:315)
    at maybeClose (internal/child_process.js:1021)
    at Socket.<anonymous> (internal/child_process.js:443)
    at Socket.emit (events.js:315)
    at Pipe.<anonymous> (net.js:674)
```

The error is correctly reported when calling `Sentry.captureException(error)` but in the Sentry web interface the stacktrace is empty. Only the first two lines appear:

```
Command failed: commandThatFails
/bin/sh: commandThatFails: command not found
```

It seems the Sentry client tries to parse the stack as a Chrome stack, but as the error comes from the Node integration, the format does not match and all frames are skipped.

timfish:
Closes #308

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "chai-subset": "^1.6.0",
     "cross-env": "^7.0.3",
     "eslint": "7.27.0",
-    "electron": "15.3.0",
+    "electron": "19.0.8",
     "@electron/get": "^1.13.0",
     "electron-mocha": "^11.0.0",
     "electron-latest-versions": "^0.1.7",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fix: Correctly parse mixed Chrome/node stack traces in the renderer (#509)
+
 ## 4.0.0-beta.1
 
 Updating the underlying Sentry JavaScript SDK's to v7 forces a major version bump due to minor breaking changes in user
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -85,7 +85,7 @@
     "chai-subset": "^1.6.0",
     "cross-env": "^7.0.3",
     "eslint": "7.27.0",
-    "electron": "15.3.0",
+    "electron": "19.0.8",
```
AbhiPrasad, line 88: Intentional?

AbhiPrasad (approved):

package.json
timfish: Yeah, this is the version used for unit tests. If we don't keep this updated we can get dependabot security alerts.

This does mean we only run the unit tests in one Electron version but the e2e tests should keep on top of any breakages between Electron versions.

PR #510, opened 2022-07-09
feat: Optionally attach screenshots

### LINKED ISSUE #505: Attach screenshots with the SDK
Now that we have mobile screenshots, let's do it for desktop! https://blog.sentry.io/2022/07/07/introducing-mobile-screenshots-and-suspect-commits/

https://develop.sentry.dev/sdk/features/#screenshots

This will have to be **opt-in**. We can configure this through an `attachScreenshot` option, and add a BIG warning about PII

timfish:
Closes #505 

- Adds `attachScreenshot` property to main process options
- Adds `Screenshots` integration which is included by default so it can be enabled via the above option
- On JavaScript error, windows are captured as PNG and added as attachments
- Modified e2e server so that it captures different types of attachments rather than just minidumps
- Modified all e2e test events to use `attachments` rather than `dumpFile`

### DIFF
--- a/src/main/sdk.ts
+++ b/src/main/sdk.ts
@@ -15,6 +15,7 @@ import {
   Net,
   OnUncaughtException,
   PreloadInjection,
+  Screenshots,
   SentryMinidump,
 } from './integrations';
 import { configureIPC } from './ipc';
@@ -30,6 +31,7 @@ export const defaultIntegrations: Integration[] = [
   new OnUncaughtException(),
   new PreloadInjection(),
   new AdditionalContext(),
+  new Screenshots(),
   ...defaultNodeIntegrations.filter((integration) => integration.name !== 'OnUncaughtException'),
 ];
 
@@ -62,6 +64,11 @@ export interface ElectronMainOptionsInternal extends Options<ElectronOfflineTran
    * scheme is used.
    */
   getRendererName?: (contents: WebContents) => string | undefined;
+
+  /**
+   * If set to true, screenshots will be captured and included for all events.
+   */
+  attachScreenshot?: boolean;
 }
 
 // getSessions and ipcMode properties are optional because they have defaults
--- a/src/main/sdk.ts
+++ b/src/main/sdk.ts
@@ -68,7 +68,9 @@ export interface ElectronMainOptionsInternal extends Options<ElectronOfflineTran
   /**
    * Screenshots may contain PII and is an opt-in feature
    *
-   * If set to true, screenshots will be captured and included for all events.
+   * If set to true, screenshots will be captured and included with all JavaScript events.
+   * Screenshots are not included for native crashes since it's not possible to capture images of crashed Electron
+   * renderers.
    */
   attachScreenshot?: boolean;
 }
--- a/src/main/sdk.ts
+++ b/src/main/sdk.ts
@@ -66,6 +66,8 @@ export interface ElectronMainOptionsInternal extends Options<ElectronOfflineTran
   getRendererName?: (contents: WebContents) => string | undefined;
 
   /**
+   * Screenshots may contain PII and is an opt-in feature
+   *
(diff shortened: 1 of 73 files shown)


AbhiPrasad:
Woah - amazing work Tim! Will review on Monday. Is there a example event on sentry.io with this - would like to see what it looks like.

timfish:
Sorry my test app wasn't more interesting 🤣

image

image

timfish:
Some other things that might be worth considering now or later:
- Should we only capture windows that are visible? 
- Should the focussed window be the first screenshot?
- Should we try and track which renderer the error came from and make that the first screenshot?
- For main process errors, is capturing window screenshots that useful?

AbhiPrasad:
> Should we only capture windows that are visible?

My gut instinct here is yes, as we want to capture what a user is seeing. Are there any possible scenarios where it might be useful to have it for non-visible windows?

> Should the focused window be the first screenshot?

This feels like the correct behaviour to me.

> Should we try and track which renderer the error came from and make that the first screenshot?

Why would we do this? Feels like un-needed complication. 

> For main process errors, is capturing window screenshots that useful?

I think we can stick with capturing screenshots for every single error, and just iterate on feedback.

> Should we scale the screenshots so they're not full size?

As per the [spec](https://develop.sentry.dev/sdk/features/#screenshots): `Image size, if possible should stay below 2 MB but quality/size could be configurable`. I think we should only adjust to make sure we stay within the size limit, otherwise we should not scale for now. We can revisit!

src/main/sdk.ts
```diff
@@ -62,6 +64,13 @@ export interface ElectronMainOptionsInternal extends Options<ElectronOfflineTran
    * scheme is used.
    */
   getRendererName?: (contents: WebContents) => string | undefined;
+
+  /**
+   * Screenshots may contain PII and is an opt-in feature
+   *
+   * If set to true, screenshots will be captured and included for all events.
```
AbhiPrasad, line 71: Leave a note that we don't capture for native crashes?

timfish:
>> Should we try and track which renderer the error came from and make that the first screenshot?

> Why would we do this? Feels like un-needed complication.

I would be looking at screenshots to see what the user was doing when the error occurred. I can imagine getting an event with multiple screenshots, one for each window and wanting to know which one is most relevant for the error I'm looking at. 

>> Should the focused window be the first screenshot?

> This feels like the correct behaviour to me.

Defaulting to the focussed window as the first screenshot could even suggest the error came from there.

In an ideal world we would tag the screenshots with some context like focus, and "errored", much like stack trace envelope API allows you to set which thread crashed. Could we put some of this into the file names?

`screenshot-focus.png`, `screenshot-error.png`, `screenshot-focus-error.png`, etc

> I think we can stick with capturing screenshots for every single error, and just iterate on feedback.

Yep, it's probably good enough as it is for now!

AbhiPrasad:
> I would be looking at screenshots to see what the user was doing when the error occurred. I can imagine getting an event with multiple screenshots, one for each window and wanting to know which one is most relevant for the error I'm looking at.

Good point - it prevents you from looking back and forth.

> In an ideal world we would tag the screenshots with some context like focus, and "errored", much like stack trace envelope API allows you to set which thread crashed. Could we put some of this into the file names?

At the current moment it wouldn't do anything (as we require the strict naming scheme to display in the product accordingly) but good suggestion. 

I will raise these two points at our next internal SDK sync meeting, see what people think - and bring that feedback back here. In the mean time, I think we can ship with what we have!

AbhiPrasad:
We just met and chatted - and the solution of using `event.contexts` was brought up.

So we would add an event context (maybe keyed under `screenshot`) that would have the information about what screenshots were focused or errored or anything else.

We might have to first make a PR against https://github.com/getsentry/develop/blob/master/src/docs/sdk/features.mdx#screenshots describing our ideal behaviour - and then inform the other teams working on this.

PR #516, opened 2022-07-16
fix: Check preload path is absolute

### LINKED ISSUE #491: preload script must have absolute path
## Versions + Platform

- [x] Electron version - `electron@v18.2.0`
- [x] Platform - `macOS`

## Description

There are lots of error messages coming from the main process:

```
[22375:0526/171429.459534:ERROR:session_preferences.cc(38)] preload script must have absolute path: ../../common/temp/node_modules/.pnpm/@sentry+electron@3.0.7/node_modules/@sentry/electron/esm/preload/index.js
```

Screen Shot 2022-05-26 at 5 16 58 PM

timfish:
Closes #491

It looks like webpack can convert `require.resolve` to relative paths that Electron then complains about. 

This change ensures that we only attempt to use preload paths that came from `require.resolve`, ie. only absolute paths will be attempted.

### DIFF
--- a/src/main/integrations/preload-injection.ts
+++ b/src/main/integrations/preload-injection.ts
@@ -4,6 +4,7 @@ import { Integration } from '@sentry/types';
 import { logger } from '@sentry/utils';
 import { app } from 'electron';
 import { existsSync } from 'fs';
+import { isAbsolute } from 'path';
 
 import { IPCMode } from '../../common';
 import { rendererRequiresCrashReporterStart } from '../electron-normalize';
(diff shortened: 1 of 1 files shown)


src/main/integrations/preload-injection.ts
```diff
@@ -49,7 +50,7 @@ export class PreloadInjection implements Integration {
       //
     }
 
-    if (path && typeof path === 'string' && existsSync(path)) {
+    if (path && typeof path === 'string' && isAbsolute(path) && existsSync(path)) {
```
AbhiPrasad, line 53: Can we log out an error if the path is not absolute? Or atleast attempt to log out the path in the debug message below?

AbhiPrasad (approved):

src/main/integrations/preload-injection.ts
timfish: I would be wary of logging any warnings here since all this change actually does it stop Electron from logging warnings about preload non-absolute paths. 

If injecting preload fails, we fallback to custom protocol + `fetch` so the SDK should continue to function as normal.

AbhiPrasad: Ah ok that makes sense, thanks for the details!

PR #527, opened 2022-08-02
Update to Sentry JavaScript 7.8.1

### LINKED ISSUE #503: Issues on Windows don't get merged because of user path

## Versions + Platform

- [x] Electron version - `electron@v18.2.3`
- [x] Platform - `Windows`

## Description

In my Electron app issues are not properly merged from Windows users.
In the app I'm basically doing:

```
Sentry.captureException(new Error('Tried to send recordingEvent [...]'));
```

These are the duplicate issues that I get in Sentry:

Screenshot 2022-07-06 at 12 45 30

I can see the full user path as a description. I'd expect `C:\Users\myusername` to be replaced to e.g. `~` (or something else), and I guess then it can be merged properly? On macOS this doesn't seem to be an issue.

Or is there another way to rewrite the errors so this user path doesn't get included?

timfish:
Fixes #503

### DIFF
--- a/test/e2e/test-apps/javascript/main-error/event.json
+++ b/test/e2e/test-apps/javascript/main-error/event.json
@@ -69,6 +69,7 @@
                 "colno": 0,
                 "filename": "app:///src/main.js",
                 "function": "{{function}}",
+                "module": "main",
                 "in_app": true,
                 "lineno": 0
               }
--- a/package.json
+++ b/package.json
@@ -57,19 +57,19 @@
     "e2e": "cross-env TS_NODE_PROJECT=tsconfig.json xvfb-maybe mocha --require ts-node/register/transpile-only --retries 3 ./test/e2e/*.ts"
   },
   "dependencies": {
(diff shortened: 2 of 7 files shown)


Lms24 (approved):

AbhiPrasad:
Can we add a test that confirms the fix?

test/e2e/test-apps/javascript/main-error/event.json
```diff
@@ -69,6 +69,7 @@
                 "colno": 0,
                 "filename": "app:///src/main.js",
                 "function": "{{function}}",
+                "module": "main",
```
timfish, line 72: Added this check in here. If it passes on all platforms we're all good!

AbhiPrasad: image

(bringing the :salute_canada: emoji react to GH to share with the world :p)

AbhiPrasad (approved):

PR #535, opened 2022-08-15
Make `beforeSend` work from renderers

### LINKED ISSUE #533: Don't manage to send attachments
https://sentry.io/organizations/stephan-petzl/issues/3499972015/attachments/?project=1554144&query=is%3Aunresolved

## Versions + Platform

- SDK version - @sentry/electron@4.0.0
- Electron version - electron@12.2.3
- Platform - Mac OS

## Description

I am configuring Sentry like this:

```
function configure() {
  var consoleError = console.error
  console.error = function() {
    var args = Array.prototype.slice.call(arguments)
    const error = reduceConsoleArgs(args)
    Sentry.configureScope(scope => {
      console.log('configureScope1')
      const runtimeData = getRuntimeData(store)
      scope.addAttachment({ filename: 'context.json', data: JSON.stringify(runtimeData) })
    })
    Sentry.captureException(error, { level: 'error' })
    return consoleError.apply(console, args)
  }
}
```

Unfortunately no attachment is added to the report.
Can somebody give me a hint?

timfish:
Renderer events are passed to the main process via the `EventToMain` integration. The downside of using an integration/global event processor to do this is that the renderer `beforeSend` never gets called.

This PR calls `beforeSend` from `EventToMain` so that it functions as it should!

This closes #533.

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -1,5 +1,5 @@
 import { captureEvent, configureScope, Scope } from '@sentry/core';
-import { Event } from '@sentry/types';
+import { Event, EventHint } from '@sentry/types';
 import { logger, SentryError } from '@sentry/utils';
 import { app, ipcMain, protocol, WebContents } from 'electron';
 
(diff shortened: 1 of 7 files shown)


timfish:
🤔 just realised this doesn't take into account any modifications to `hint` which is required to add attachments from `beforeSend`...

timfish:
This now passes `[event,  hint]` serialised from render to main and adds a test that ensures that `hint.attachments` added in the renderer `beforeSend` are submitted in envelopes from the main.

My understanding is this will currently only work with `string` attachments since `UInt8Array` get's mangled by json stringify.

AbhiPrasad:
`beforeSend` explicitly is meant to run after the event processors - https://develop.sentry.dev/sdk/unified-api/#event-pipeline, not sure if we can make this change.

timfish:
> `beforeSend` explicitly is meant to run after the event processors

If `EventToMain` is the last integration, will this not be true?

timfish:
#610 means this will no longer be required

PR #548, opened 2022-09-09
Update to v7.12.1 of JavaScript SDKs

timfish:
This also disables the node `Context` integration since this clashes with existing features in the Electron SDK and it's too complicated to migrate to using `Context` for now without causing breaking changes.

### DIFF
--- a/test/e2e/test-apps/other/browser-tracing/package.json
+++ b/test/e2e/test-apps/other/browser-tracing/package.json
@@ -4,6 +4,6 @@
   "main": "src/main.js",
   "dependencies": {
     "@sentry/electron": "3.0.0",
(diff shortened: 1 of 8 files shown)


test/e2e/test-apps/other/browser-tracing/package.json
```diff
@@ -4,6 +4,6 @@
   "main": "src/main.js",
   "dependencies": {
     "@sentry/electron": "3.0.0",
-    "@sentry/tracing": "7.11.1"
+    "@sentry/tracing": "7.12.1"
```
AbhiPrasad, line 7: Is there some way we can automate this? or make the e2e tests grab the local npm package instead?

AbhiPrasad (approved):

Issue #534, opened 2022-08-15
Why use a synchronous API (writeFileSync) for file writing


niugm:
Sentry is a great project, we use it on the Electron client, and there's a question I don't quite understand: Why use a synchronous API (writeFileSync) for file writing?On some low-end devices, especially computers that use mechanical disks, the main process execution can be blocked

timfish:
Thanks for opening an issue about this.

It looks like `writeFileSync` is used in the `Store` which has gone virtually untouched for over 3 years when it was pulled from `@sentry/utils`. 

I did quickly have a go at making the store entirely async and got some test failures, so it looks like the session handling is at least partially relying on the sync writing. 

<img width="534" alt="image" src="https://user-images.githubusercontent.com/1150298/184613339-6e5e2a41-2c89-445d-b669-4ff9391800c5.png">

niugm:
> 

Thank you for your attention to this feedback, looking forward to the optimized version

AbhiPrasad:
Re-opening as fix was reverted in https://github.com/getsentry/sentry-electron/pull/552

PR #550, opened 2022-09-12
fix: Rewrite async file store

timfish:
fixes https://github.com/getsentry/sentry-electron/issues/534

I found a race condition in the `Store` after making it async (#542) where the json file would be read back as empty because it was being written at the same time. This code originally came from `@sentry/utils` but is only used here now. I decided it would be a good idea to review how it works and compare to the requirements for the Electron SDK.

The first thing that became clear after testing is that the store does not reliably flush concurrent changes to disk. If the store is already flushing, it skips flushing entirely and these changes don't make it to disk:
https://github.com/getsentry/sentry-electron/blob/055b1c432b4d25bdb099be9047cd998317ccce87/src/main/store.ts#L50-L63

It uses `setImmediate` to flush in a non-blocking way but this probably only made sense when all the `fs` calls were sync. The immediate flushing was only used for sessions where we want changes to make it to disk asap to reliably track abnormal exits. 

I stripped the store back to the bare minimum and tested. It's clear that for some usages write throttling is required to keep the number of file writes to a minimum. Flushing to disk on every scope change would result in far too many file writes.

Fully locking store read/write has the potential to impact performance in heavy read/write scenarios but this is not how the store is used. In all cases it's much more important that the data returned by the store is correct and race conditions are avoided.

### Sessions
Sessions are only read/write at app start/exit. Abnormal session detection relies on being able to read the previous session and quickly write the new session without races.

### Scope storage
Scope changes are only stored so that we have a representative scope to be sent with main process native crashes on subsequent app starts. This is write heavy and throttling could result in scope slightly out of sync but since native crashes are already decoupled from JavaScript scope capture this is less of a concern.

### Transport queue
The transport queue is used to buffer envelopes for offline support and is only used when events are captured and cannot be sent. It is important that it works correctly and doesn't drop envelopes.

This PR:
- Removes sync directory creation and fixes some bugs in `mkdirp` where the promises weren't awaited and errors caught in try/catch
- Simplifies the `Store`
  - Removes the broken `flushing` logic
  - Uses `async-mutex` to fully lock reads and writes to remove race conditions
- Adds a `BufferedWriteStore` that wraps `Store` and throttles disk writes @ 500ms by default
- Session storage uses `Store` where writes are minimal and throttling may cause missed sessions
- The transport queue and scope storage use `BufferedWriteStore` because these can see a high number of writes and we don't want every one of these to result in file io.

## `async-mutex`

I spent quite a few hours trying a create what this library already does well. 

This code has been vendored and significantly simplified!


src/main/store.ts
```diff
@@ -78,15 +77,25 @@ export class Store<T> {
    * constructor is used.
    */
   public async get(): Promise<T> {
-    if (this._data === undefined) {
-      try {
-        this._data = JSON.parse(await readFileAsync(this._path, 'utf8'), dateReviver) as T;
-      } catch (e) {
-        this._data = this._initial;
+    return this._lock.runExclusive(async () => {
```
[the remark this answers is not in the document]
timfish: Once the store has completed its initial load from file, the value is stored in `this._data` and returned pretty quickly. We need to lock specifically to stop races during the initial load. We could perhaps move the get locking to only be around the file reading. That's might be worth a try 🤔

Reading from the store doesn't happen that much:
- Sessions - Only reads once at startup to check for previous un-completed session
- Scope - Only reads at startup if there are unreported minidumps from previous session
- Transport queue - Only reads to insert if you're offline and while there are still items in the queue to retry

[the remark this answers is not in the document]
lforst: If it's just during initial load or happening rarely, a mutex seems totally fine to me!

timfish:
It's also worth nothing that in this PR I make both `async-mutex` and `deepmerge` use a fixed versions in `package.json`.

If we're using dependencies and we've fully tested them working for our use case, using a fixed version avoids breakages from patch versions that make semver breaking changes and also avoids supply-chain attacks via these dependencies!

src/main/store.ts
```diff
@@ -78,50 +76,84 @@ export class Store<T> {
...
+    this._throttleTime = throttleTime;
+  }
+
+  /** @inheritdoc */
+  public override async set(data: T): Promise<void> {
+    this._data = data;
+
+    this._pendingWrite = {
+      // Overwrite with the latest data
+      data,
```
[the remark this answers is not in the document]
timfish: This allows `this. _pendingWrite` to be updated by a subsequent `set` that occurs before the changes are written to disk. If we pass the value down through `_writePending` we'll need to queue up another write if there are any changes before flushing.

package.json
```diff
@@ -63,7 +63,8 @@
     "@sentry/node": "7.12.1",
     "@sentry/types": "7.12.1",
     "@sentry/utils": "7.12.1",
-    "deepmerge": "^4.2.2",
+    "async-mutex":"0.4.0",
```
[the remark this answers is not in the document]
timfish: Ok, will vendor this and trim it down!

timfish:
This closes #534

src/common/mutex.ts
```diff
@@ -0,0 +1,79 @@
+// This code was originally copied from  https://github.com/DirtyHairy/async-mutex
```
AbhiPrasad, line 1: Can we include the commit sha + the license of the library here?

src/main/store.ts
```diff
@@ -78,50 +76,84 @@ export class Store<T> {
...
-      // system, or problems with network drives. We cannot fix or handle this
-      // issue and must resume gracefully. Thus, we have to ignore this error.
-    } finally {
-      this._flushing = false;
+/**
+ * Extends Store to throttle writes.
+ */
+export class ThrottledStore<T> extends Store<T> {
+  /** The minimum time between writes */
+  private readonly _throttleTime?: number;
```
AbhiPrasad, line 121: Essentially, we are using the timer to dedupe batched updates right? Can we just call this `DedupedStore` then?

timfish: Deduping is "the process of removing identical entries" which is not really what's happening here. 

We are throttling or buffering flushes to disk. 

How about `BufferedWriteStore`?

AbhiPrasad: Yes you're right - and `BufferedWriteStore` maybe fits better here. 

I'm just thinking if there's a way we can avoid the 500ms and just have a queue + event loop take care of this for us, having the throttle feels like kind of a code smell to me.

AbhiPrasad: > There's no point flushing values to disk if you already know they'll be overwritten

Great point, this convinced me. Let's do this for now and improve later if we get feedback.

AbhiPrasad (approved):

PR #551, opened 2022-09-12
fix: Page titles in breadcrumbs should not change

CliftonH:
We noticed an issue where page titles were showing up in breadcrumbs and discovered that the breadcrumbs' `data` properties are references to `RENDERERS` object [here](https://github.com/getsentry/sentry-electron/blob/ef1839a2b2268909ab8f2209dc5f8a5c6fd1fec4/src/main/renderers.ts#L12). Thus, if the page title changes after a breadcrumb is created but before it is sent with an event, [this handler](https://github.com/getsentry/sentry-electron/blob/ef1839a2b2268909ab8f2209dc5f8a5c6fd1fec4/src/main/renderers.ts#L43) will also change the `data` property of the breadcrumb.

This change prevents that by setting the `data` property of the breadcrumb to a clone of the renderer state. Also, this change prevents the breadcrumb integration from modifying the renderer state when writing its `data` prop.

### DIFF
--- a/src/main/integrations/electron-breadcrumbs.ts
+++ b/src/main/integrations/electron-breadcrumbs.ts
@@ -183,14 +183,10 @@ export class ElectronBreadcrumbs implements Integration {
         };
 
         if (id) {
-          const state = getRendererProperties(id);
+          breadcrumb.data = { ...getRendererProperties(id) };
 
-          if (!this._options.captureWindowTitles && state?.title) {
-            delete state.title;
-          }
-
-          if (state) {
-            breadcrumb.data = state;
+          if (!this._options.captureWindowTitles && breadcrumb.data?.title) {
+            delete breadcrumb.data?.title;
           }
         }
 


src/main/integrations/electron-breadcrumbs.ts
```diff
@@ -183,14 +183,10 @@ export class ElectronBreadcrumbs implements Integration {
         };
 
         if (id) {
-          const state = getRendererProperties(id);
+          breadcrumb.data = { ...getRendererProperties(id) };
```
MarshallOfSound, line 186: `getRendererProperties` could return `null`

CliftonH: Unless I'm mistaken, `getRendererProperties` should only ever return some object or `undefined`, which this change should handle. Am I overlooking something?

MarshallOfSound: Oh interesting, I didn't think `...null` was valid JS 🤔 Apparently it is?

AbhiPrasad: Yeah this should be fine to cover the case here.

AbhiPrasad (approved):
Thanks for the patch! Took the liberty of editing the title to match our conventions. We can try to get this released tomorrow, @timfish does that work for you?

timfish:
To release we either need to revert #542 or merge #550 since I found a data race in the async file store changes. 
Reverting #542 is certainly easier!

lobsterkatie:
@timfish - We've got some folks asking if we could write a test for this fix, to make sure there's never a regression in the future. Is that something you might be able to help with?

cc @yuval-sentry

timfish:
Yes, I will add some tests!

yuval-sentry:
Thank you!!!

lobsterkatie:
Thanks, @timfish!

PR #552, opened 2022-09-12
Revert "feat: Make the store completely async"

timfish:
Reverts getsentry/sentry-electron#542

This resulted in data races for file read/write


AbhiPrasad (approved):
Sounds good, we can try to do the async refactor in another release

for others, this helps unblock https://github.com/getsentry/sentry-electron/pull/551

PR #574, opened 2022-10-12
ref(ci): fix set-output / set-state deprecation

asottile-sentry:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Committed via https://github.com/asottile/all-repos


AbhiPrasad (approved):
Thanks!

PR #575, opened 2022-10-13
feat: Support passing JavaScript framework specific SDK init as second init parameter

### LINKED ISSUE #506: What is the right way to setup Sentry Electron SDK with Vite & Vue 3
## Versions + Platform

- [x] Electron version - `electron@v19.0.8`
- [x] Platform - `macOS`

## Description

I am using `@sentry/vue` package to collect exceptions from renderer process, it is not very straightforward and I have to revise the Sentry event before sending it to server, otherwise source mapping doesn't work.

```
import * as Sentry from '@sentry/vue';
Sentry.init({
  app,
  dsn: __SENTRY_DSN__,
  release: __SENTRY_RELEASE__,
  beforeSend: (event: Sentry.Event, hint: Sentry.EventHint) => {

    // update the stacktrace before sending it to Sentry, this will strip out the asar path - i.e. those between two brackets:
    // file:///[myapp.app/Contents/Resources/app.asar/]packages/renderer/dist/index.f3984842.js

    if (event.exception?.values) {
      event.exception.values = event.exception.values.map(e => {
        if (e.stacktrace?.frames) {
          e.stacktrace.frames = e.stacktrace.frames.map(f => {
            if (f.filename) {
              f.filename = f.filename.replace(/.*app\.asar/, 'app://');
            }
            return f;
          });
        }
        return e;
      });
    }
    return event;
  },
});
```

It works with this approach but I would like to check if there is an easier / better way available. I firstly tried `@sentry/electron` package but it doesn't capture exceptions in renderer process for me, maybe it is because I am using Vue 3 and not vanilla.

My other question is, it seems the exceptions in renderer process won't be printed in console anymore once they are reported to Sentry, is this an intended behavior?

timfish:
This copies the approach used in [`@sentry/capacitor`](https://github.com/getsentry/sentry-capacitor/blob/6aeb492f540f468f5190c025ebd799fdb7dd3e8e/src/sdk.ts#L20-L23). 

```ts
import * as Sentry from "@sentry/capacitor";
import { init as sentryAngularInit }  from "@sentry/angular";

// Init by passing the sibling SDK's init as the second parameter.
Sentry.init({
  dsn: "__DSN__",
}, sentryAngularInit);
```

This involves minimal code changes and is not a breaking change since to the 2nd parameter has a default of the browser init!

I've extended the react and vite example apps to use this approach.

Closes #506

### DIFF
--- a/test/e2e/recipe/index.ts
+++ b/test/e2e/recipe/index.ts
@@ -15,7 +15,7 @@ import { parseRecipe, TestRecipe } from './parser';
 export * from './normalize';
 
 const SENTRY_KEY = '37f8a2ee37c0409d8970bc7559c7c7e4';
-const TRACING_VERSION = require('../../../package.json').devDependencies['@sentry/tracing'];
+const JS_VERSION = require('../../../package.json').dependencies['@sentry/core'];
 
 const log = createLogger('Recipe Runner');
 
@@ -121,8 +121,10 @@ export class RecipeRunner {
             /"@sentry\/electron": ".*"/,
             `"@sentry/electron": "file:./../../../../sentry-electron-${SDK_VERSION}.tgz"`,
           )
-          // We replace the @sentry/tracing dependency version to match that of @sentry/electron
-          .replace(/"@sentry\/tracing": ".*"/, `"@sentry/tracing": "${TRACING_VERSION}"`);
+          // We replace the Sentry JavaScript dependency versions to match that of @sentry/electron
+          .replace(/"@sentry\/tracing": ".*"/, `"@sentry/tracing": "${JS_VERSION}"`)
+          .replace(/"@sentry\/react": ".*"/, `"@sentry/react": "${JS_VERSION}"`)
+          .replace(/"@sentry\/vue": ".*"/, `"@sentry/vue": "${JS_VERSION}"`);
       }
 
       writeFileSync(path, content);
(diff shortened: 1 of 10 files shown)


test/e2e/recipe/index.ts
```diff
@@ -121,8 +121,10 @@ export class RecipeRunner {
             /"@sentry\/electron": ".*"/,
             `"@sentry/electron": "file:./../../../../sentry-electron-${SDK_VERSION}.tgz"`,
           )
-          // We replace the @sentry/tracing dependency version to match that of @sentry/electron
-          .replace(/"@sentry\/tracing": ".*"/, `"@sentry/tracing": "${TRACING_VERSION}"`);
+          // We replace the Sentry JavaScript dependency versions to match that of @sentry/electron
+          .replace(/"@sentry\/tracing": ".*"/, `"@sentry/tracing": "${JS_VERSION}"`)
+          .replace(/"@sentry\/react": ".*"/, `"@sentry/react": "${JS_VERSION}"`)
+          .replace(/"@sentry\/vue": ".*"/, `"@sentry/vue": "${JS_VERSION}"`);
```
AbhiPrasad, line 127: there is also `@sentry/svelte`, `@sentry/ember`, and `@sentry/angular`!

AbhiPrasad (approved):
Great change! Reminder to also update the sentry docs!

test/e2e/recipe/index.ts
timfish: Indeed, although we don't have any example tests for those... yet 😁

I'm also aware that the examples don't actually tests the vue/react SDK functionality!

PR #576, opened 2022-10-13
fix: Include `sentry_key` in ping URL so it does not end up in breadcrumbs

timfish:
When using `fetch` to communicate with the main process, we include `sentry_key` in the URL so that it does not result in a fetch breadcrumb.

It appears this was left off of the `ping` "command" resulting in a breadcrumb!

### DIFF
--- a/src/renderer/ipc.ts
+++ b/src/renderer/ipc.ts
@@ -13,7 +13,7 @@ function getImplementation(): IPCInterface {
 
   logger.log('IPC was not configured in preload script, falling back to custom protocol and fetch');
 
-  fetch(`${PROTOCOL_SCHEME}://${IPCChannel.PING}`).catch(() =>
+  fetch(`${PROTOCOL_SCHEME}://${IPCChannel.PING}/sentry_key`).catch(() =>
     console.error(`Sentry SDK failed to establish connection with the Electron main process.
  - Ensure you have initialized the SDK in the main process
  - If your renderers use custom sessions, be sure to set 'getSessions' in the main process options


src/renderer/ipc.ts
```diff
@@ -13,7 +13,7 @@ function getImplementation(): IPCInterface {
 
   logger.log('IPC was not configured in preload script, falling back to custom protocol and fetch');
 
-  fetch(`${PROTOCOL_SCHEME}://${IPCChannel.PING}`).catch(() =>
+  fetch(`${PROTOCOL_SCHEME}://${IPCChannel.PING}/sentry_key`).catch(() =>
```
AbhiPrasad, line 16: Shouldn’t this be a query string and not part of the path?

timfish: It doesn't matter since the `fetch` breadcrumb checking [matches the entire url](https://cs.github.com/getsentry/sentry-javascript/blob/d64550effb035d069a956e1fdc1b45ba51d289fc/packages/browser/src/integrations/breadcrumbs.ts#L207) and in the main process, the protocol handler returns an empty response for anything that isn't `scope` or `event`.

timfish: We already use this to exclude `event` and `scope` messages but it was left off `ping`!

https://github.com/getsentry/sentry-electron/blob/ffad61d621119eee8f6d42e361e393fcdb478905/src/renderer/ipc.ts#L27

AbhiPrasad: Ahhhh I see, thanks for clarifying.

AbhiPrasad (approved):

PR #580, opened 2022-10-29
feat: Add additional options and queue status callback to offline transport

timfish:
This PR adds some options to the offline transport to allow more configuration and notification of the current request queue length. 

- Adds `maxQueueAgeDays` and `maxQueueCount` options which simply get passed down to `PersistedRequestQueue`
- Modifies the `add` and `pop` functions in `PersistedRequestQueue` to return the current length of the queue
- Adds a `queuedLengthChanged` function to the offline transport options
- Calls `queuedLengthChanged` whenever the length of the request queue changes

The existing `beforeSend` option allows developers to override the queuing behaviour and queue events until users have consented to sending. 
The addition of the `queuedLengthChanged` option allows prompting users to consent or connect to the internet so they can be sent.


timfish:
#581 should fix the test failures

AbhiPrasad (approved):

PR #586, opened 2022-11-15
fix: Attempt to send minidumps for all `process-gone` reasons

timfish:
It was reported that `webContents.forcefullyCrashRenderer()` results in a minidump being generated but `render-process-gone` notifies that the process was `killed`.

We were ignoring all reasons which were not crashes and this resulted in ignoring some minidumps and then sending them on next app start and assuming they came from the main process.

This PR:
- Looks for minidumps for all exit reasons
- Includes the `exit.reason` in tags so customers can filter them

### DIFF
--- a/src/main/integrations/sentry-minidump/index.ts
+++ b/src/main/integrations/sentry-minidump/index.ts
@@ -6,7 +6,7 @@ import { app, crashReporter } from 'electron';
 
 import { mergeEvents } from '../../../common';
 import { getEventDefaults } from '../../context';
-import { CRASH_REASONS, onChildProcessGone, onRendererProcessGone } from '../../electron-normalize';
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):

Cefold:
After `webContents.forcefullyCrashRenderer()` event, there was a minidump under `pending` folder, but it seems that Sentry cannot find it. It will upload it until next app restart, thus `event.process` will be classified as `browser` instead of `renderer`.
 
```
Sentry Logger [log]: 'renderer' process 'killed'
Sentry Logger [log]: Found 0 minidumps
```

timfish:
> but it seems that Sentry cannot find it

It looks like the issue is that the minidump does not exist when `render-process-gone` is called. I need to do some more testing!

PR #592, opened 2022-11-18
fix: Also check crashpad `pending` directory on macOS when renderer process exits

### LINKED ISSUE #591: Minidump is not picked up after `webContents.forcefullyCrashRenderer()`
As per [this comment](https://github.com/getsentry/sentry-electron/pull/586#issuecomment-1315716418), cc @Cefold

> After `webContents.forcefullyCrashRenderer()` event, there was a minidump under pending folder, but it seems that Sentry cannot find it. It will upload it until next app restart, thus event.process will be classified as browser instead of renderer.

It looks like this is caused by the fact that the minidump ends up in the `pending` directory until the app restarts on macOS.

timfish:
Closes #591

On macOS, if you call `webContents.forcefullyCrashRenderer()` the resulting minidump ends up in the `pending` subdirectory and doesn't get copied to completed until the app restarts. This causes it to be classed as a main process crash.

This PR means we now check the `pending` directory for minidumps on macOS and adds a test that ensures minidumps are correctly handled when caused by `webContents.forcefullyCrashRenderer()`.

### DIFF
--- a/src/main/integrations/sentry-minidump/minidump-loader.ts
+++ b/src/main/integrations/sentry-minidump/minidump-loader.ts
@@ -94,6 +94,7 @@ function crashpadMinidumpLoader(): MinidumpLoader {
     await deleteCrashpadMetadataFile(crashesDirectory).catch((error) => logger.error(error));
 
     const dumpDirectory = join(crashesDirectory, crashpadSubDirectory);
+    const pendingDirectory = join(crashesDirectory, 'pending');
 
     return (await readDirAsync(dumpDirectory))
       .filter((file) => file.endsWith('.dmp'))
@@ -104,7 +105,19 @@ function crashpadMinidumpLoader(): MinidumpLoader {
           path,
           load: () => readFileAsync(path),
         };
-      });
+      })
+      .concat(
+        (await readDirAsync(pendingDirectory))
+          .filter((file) => file.endsWith('.dmp'))
+          .map((file) => {
+            const path = join(pendingDirectory, file);
+
+            return {
+              path,
+              load: () => readFileAsync(path),
+            };
+          }),
+      );
   });
 }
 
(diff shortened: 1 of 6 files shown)


src/main/integrations/sentry-minidump/minidump-loader.ts
```diff
@@ -86,20 +86,35 @@ async function deleteCrashpadMetadataFile(crashesDirectory: string, waitMs: numb
...
+    dumpDirectories.push(join(crashesDirectory, 'pending'));
+  }
+
   return createMinidumpLoader(async () => {
     await deleteCrashpadMetadataFile(crashesDirectory).catch((error) => logger.error(error));
 
-    const dumpDirectory = join(crashesDirectory, crashpadSubDirectory);
-
-    return (await readDirAsync(dumpDirectory))
+    return (await readDirsAsync(dumpDirectories))
```
AbhiPrasad, line 115: l: Can we extract this out into a variable, little cleaner to understand.

```suggestion
    const loader = await readDirsAsync(dumpDirectories);
    return loader
```

src/main/integrations/sentry-minidump/minidump-loader.ts
```diff
@@ -86,20 +86,35 @@ async function deleteCrashpadMetadataFile(crashesDirectory: string, waitMs: numb
...
 }
 
+async function readDirsAsync(paths: string[]): Promise<string[]> {
+  const found: string[] = [];
+  for (const path of paths) {
+    try {
+      const files = await readDirAsync(path);
+      found.push(...files.map((file) => join(path, file)));
+    } catch (_) {
+      //
```
AbhiPrasad, line 96: should we debug log out something here?

AbhiPrasad (approved):

PR #594, opened 2022-11-19
test: Add test to ensure window titles don't end up in breadcrumbs when disabled

timfish:
This PR adds a test to ensure that window titles do not end up in breadcrumbs when this feature is disabled. This is to ensure there is no regression in the behaviour fixed by #551.

The first commit in this PR actually reverses the change in #551 to demonstrate the failing test.

A subsequent commit will reapply #551.

### DIFF
--- a/test/e2e/test-apps/other/window-titles/package.json
+++ b/test/e2e/test-apps/other/window-titles/package.json
@@ -1,0 +1,8 @@
+{
+  "name": "window-titles",
+  "version": "1.0.0",
+  "main": "src/main.js",
+  "dependencies": {
+    "@sentry/electron": "3.0.0"
+  }
+}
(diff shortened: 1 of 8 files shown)


test/e2e/test-apps/other/window-titles/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "window-titles",
+  "version": "1.0.0",
+  "main": "src/main.js",
```
AbhiPrasad, line 4: What could be kinda cool is that we use https://docs.npmjs.com/cli/v9/configuring-npm/package-json#homepage to link to the issue that this is a regression test for, but just a nit.

AbhiPrasad (approved):
Thanks for the regression test!

timfish:
I'm going to add a couple more checks to this test to ensure we're actually getting the breadcrumbs that should not contain titles!

AbhiPrasad (approved):

PR #598, opened 2022-11-30
Fix: Electron e2e version lookup

timfish:
Due to https://github.com/electron/releases/issues/188, the URL to lookup the latest versions of Electron has changed.

This PR updates `electron-latest-versions` to `0.2.0` which uses the new URL and version parsing.


AbhiPrasad (approved):

AbhiPrasad:
Should we dependabot this?

timfish:
> Should we dependabot this?

There's already [a daily workflow](https://github.com/getsentry/sentry-electron/blob/master/.github/workflows/new-versions-pr.yml) that keeps these versions updated.

PR #609, opened 2022-12-18
CI: Use volta-cli/action to use the correct node version

timfish:
This PR 
- Changes the CI scripts to use `volta-cli/action` so the node version only needs setting in a single place
- Updates the node version to v18
- Catches an error that now appears to be occasionally thrown when clearing files after e2e tests


test/e2e/context.ts
```diff
@@ -179,7 +179,11 @@ export class TestContext {
 
   private _clearAppUserData() {
     for (const dir of getDeleteDirectories(this._appName)) {
-      rmSync(dir, { recursive: true, force: true });
+      try {
+        rmSync(dir, { recursive: true, force: true });
+      } catch (_) {
+        //
+      }
```
timfish, line 186: `rmSync` appears to throw in node v18 where it didn't in v16 🤷‍♂️

lforst (approved):

Issue #606, opened 2022-12-15
Support for Session Replay


bruno-garcia:
### Problem Statement

Allow reproducing bugs through session replay

### Solution Brainstorm

Add support to Session Replay similar to how it's done for Web: https://github.com/getsentry/sentry-javascript/tree/master/packages/replay#readme

AbhiPrasad:
@timfish I gave this a try before hand and was able to get the replay's sending. I think what we have to figure out here is if [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) will work on the renderer instances as expected.

timfish:
Blockers so far: 

- Electron renderers can be detected as node if `nodeIntegration` is enabled which cause the `Replay` integration to bail out early:
```ts
  setupOnce(): void {
    if (!isBrowser()) {
      return;
    }
```

- The default Electron renderer `EventToMain` integration which passes events through the main process ends up being called and returns `null` before the Replay event processor gets called
  - Adding ordering for integrations is a fudge. More/better hooks might help here
  - Events could be passed to the main process by a transport rather than an integration but this would require parsing envelopes in the main process

- https://github.com/getsentry/sentry-javascript/pull/6576

- Error ids and replays are not getting linked up

- Replay events contain urls that should be normalised so they don't contain full paths with usernames in them
  - Passing all envelopes through the main process might be the only sensible way

PR #610, opened 2022-12-20
feat: Use a transport to send events to main process rather than via integration

### LINKED ISSUE #533: Don't manage to send attachments

timfish:
Closes #533

Also required for full Replay support (#606)

Events are sent from renderers to the main process so they can be enhanced with more context before being sent to Sentry. Previously this used the `EventToMain` integration but using an integration to do this causes issues. 

`defaultIntegration`s are always called first so `EventToMain` stopped `beforeSend` and event processors from being called in renderers.

Replay sends envelopes directly to the transport and although we could allow it to send them via the renderer (browser) transport, they contain file path URLs which likely contain local usernames (PII). If we send envelopes via the main process URLs like this can be normalised.

This PR:
- Removes `EventToMain` from the default renderer integrations
- Adds another method/route to IPC to allow sending of envelopes from renderer to main
- Adds a new default transport to renderers to send the envelopes via IPC
- Parses the envelopes in the main process and modifies or forwards to the transport as required 
- Test changes
  - Some e2e tests required their `mechanism.type` updating to `instrument` which feels like up upgrade over `generic`?
  - The test server + new `parseEnvelope` resulted in some nasty Buffer -> String -> Buffer conversion which caused malformed envelopes and needed fixing
  - The e2e tests are now forced to use Electron arm64 on M1 macs since Crashpad doesn't work in Rosetta  

### Breaking Change?
Is this a major version bump breaking change? I think probably not...

The `EventToMain` integration remains for now, it's just not in the defaults. If customers are manually adding it to integrations it'll continue to work as before and the new transport wont be used.

We only add the new default transport in renderers if a transport hasn't been defined.

If this doesn't result in a new major version. On the next major:
- Remove `EventToMain`
- Remove `sentEvent` IPC channel/route


test/e2e/download.ts
```diff
@@ -1,8 +1,18 @@
...
 import { existsSync } from 'fs';
 import { join } from 'path';
 
+function isMacArm64(): boolean {
+  if (process.platform !== 'darwin') {
+    return false;
+  }
+
+  const output = spawnSync('sysctl', ['hw.cputype']).output?.toString();
+  return output?.includes('16777228');
```
lforst, line 13: Can we add a quick comment what this is about?

test/e2e/server/index.ts
```diff
@@ -39,6 +40,27 @@ export interface TestServerEvent<T = unknown> {
...
+  return new Promise((resolve, reject) => {
+    const buf: Buffer[] = [];
+    stream.on('data', (chunk) => buf.push(chunk));
+    stream.on('end', () => resolve(Buffer.concat(buf)));
+    stream.on('error', (err) => reject(err));
+  });
+}
+
+async function getRequestBody(
+  ctx: Koa.ParameterizedContext<any, Router.IRouterParamContext<any, any>, any>,
```
lforst, line 53: l: Is there any reason we're providing `any` as generics here or could we just use the default `Koa.ParameterizedContext` without explicitly passing generics?

src/renderer/transport.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Creates a Transport that passes envelopes to the Electron main process.
+ */
+export function makeRendererTransport(options: BaseTransportOptions): Transport {
+  const ipc = getIPC();
+
+  return createTransport(options, async (request: TransportRequest): Promise<TransportMakeRequestResponse> => {
+    ipc.sendEnvelope(request.body);
+    return { statusCode: 200 };
```
lforst, line 14: Should we add a comment here stating why it is fine to always return 200 here? I am guessing because it's just transferring the event over to the main process which will then take care of rate limiting itself.

src/main/ipc.ts
```diff
@@ -106,6 +161,7 @@ function configureProtocol(options: ElectronMainOptionsInternal): void {
 function configureClassic(options: ElectronMainOptionsInternal): void {
   ipcMain.on(IPCChannel.EVENT, ({ sender }, jsonEvent: string) => handleEvent(options, jsonEvent, sender));
   ipcMain.on(IPCChannel.SCOPE, (_, jsonScope: string) => handleScope(options, jsonScope));
+  ipcMain.on(IPCChannel.ENVELOPE, ({ sender }, env: Uint8Array | string) => handleEnvelope(options, env, sender));
```
lforst, line 164: Is there a possibility that users will listen on the `IPCChannel.EVENT` themselves to do some stuff but since the `EventToMain` integration is gone by default the channel will not receive anything anymore?

I think I am fine with that being the case, however we should probably mention this in the changelog somewhere.

test/e2e/test-apps/other/scope-breadcrumbs/src/index.html
```diff
@@ -9,6 +9,10 @@
 
       init({
         debug: true,
+        beforeSend: (event, hint) => {
+          hint.attachments = [{ filename: 'screenshot.png', data: 'captureScreen()' }, ...(hint.attachments || [])];
```
lforst, line 13: I assume this is just an extra assertion you snuck into this PR? :D

lforst (approved):

src/main/ipc.ts
timfish: We don't export `IPCChannel` so to listen for this event they'd need find and use the correct string (`sentry-electron.event`). 

I suppose our IPC usage is an internal, undocumented API.

lforst: Ok right. Then it probably doesn't matter at all.

test/e2e/test-apps/other/scope-breadcrumbs/src/index.html
timfish: Before you could add attachments from the renderer process by adding them to the scope which gets synced to the main process. 

This PR fixes the ability to add envelopes via `beforeSend` too.

So this extra check is to ensure that envelopes from both routes make it into the final envelope with neither getting overwritten by the other or any other shenanigans.

PR #612, opened 2022-12-30
fix: Use `POST` for main process ping so it does not result in fetch breadcrumb

timfish:
While debugging e2e tests I noticed we still capture a breadcrumb for the ping IPC command.

```ts
      {
        timestamp: 1671415619.64,
        category: 'fetch',
        data: {
          method: 'GET',
          url: 'sentry-ipc://sentry-electron.ping/sentry_key',
          status_code: 200
        },
        type: 'http'
      }
```
This is because the `@sentry/browser` code that ignores urls with `sentry_key` in them only matches for `POST` requests.

### DIFF
--- a/src/renderer/ipc.ts
+++ b/src/renderer/ipc.ts
@@ -42,7 +42,7 @@ let cachedInterface: IPCInterface | undefined;
  * Renderer IPC interface
  *
  * Favours IPC if its been exposed via a preload script but will
- * fallback to custom protocol and fetch is IPC is not available
+ * fallback to custom protocol and fetch if IPC is not available
  */
 export function getIPC(): IPCInterface {
   if (!cachedInterface) {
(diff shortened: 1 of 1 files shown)


src/renderer/ipc.ts
```diff
@@ -13,7 +13,7 @@ function getImplementation(): IPCInterface {
 
   logger.log('IPC was not configured in preload script, falling back to custom protocol and fetch');
 
-  fetch(`${PROTOCOL_SCHEME}://${IPCChannel.PING}/sentry_key`).catch(() =>
+  fetch(`${PROTOCOL_SCHEME}://${IPCChannel.PING}/sentry_key`, { method: 'POST', body: '' }).catch(() =>
```
lforst, line 16: Just a thought: Alternatively we could filter for this request in `beforeBreadcrumb` but I think this is also fine.

lforst (approved):

src/renderer/ipc.ts
timfish: I would love integrations to [have more hooks](https://github.com/getsentry/sentry-javascript/issues/4790#issuecomment-1357818411) like for `beforeBreadcrumb`!

PR #622, opened 2023-01-11
test: Don't run sandbox tests on Linux for Electron < 13

timfish:
For release builds we test all supported Electron versions. 

It [appears](https://github.com/getsentry/sentry-electron/actions/runs/3892651247/jobs/6644902509) that tests involving sandboxed renderers no longer pass on `ubuntu-latest` for older, unsupported versions of Electron.

This PR skips those failing tests.


AbhiPrasad (approved):

AbhiPrasad:
This might be due to OpenSSL 3.0 upgrade, but disabling it sounds good to me.

PR #625, opened 2023-01-15
CI: A few CI updates and fixes

timfish:
- Update `actions/checkout` version to fix node v12 warnings
- Fix indentation 
- Remove redundant cache

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


AbhiPrasad (approved):

.github/workflows/build.yml
```diff
@@ -71,13 +71,8 @@ jobs:
     env:
       ELECTRON_VERSION: ${{ matrix.electron }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: volta-cli/action@v3
-      - uses: actions/cache@v2
```
timfish, line 76: This caches downloaded Electron binary files. This won't be much faster than just downloading them again while using up a lot of the cache allowance

PR #652, opened 2023-03-11
feat: Add callback option for more control over screenshot capture

### LINKED ISSUE #648: Only capture screenshot on certain events
### Problem Statement

I would like to request a feature where screenshot of the Electron app can be included when certain conditions match, for example, only include screenshot with `captureException`, don't upload screenshot when using `captureMessage`. Is it possible?

### Solution Brainstorm

It might be something upstream in Javascript SDK? My quick thought is changing `attachScreenshot` from a `boolean` to a `boolean | function`, and in the function we can define our filters.

timfish:
Closes #648

Extends the `attachScreenshot` option to allow a callback to be passed which when supplied is called before capturing a screenshot. The callback is passed the event and hint and returning true results in screenshot attachment.

### DIFF
--- a/src/main/integrations/screenshots.ts
+++ b/src/main/integrations/screenshots.ts
@@ -16,12 +16,18 @@ export class Screenshots implements Integration {
 
   /** @inheritDoc */
   public setupOnce(addGlobalEventProcessor: (callback: EventProcessor) => void): void {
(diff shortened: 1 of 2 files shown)


AbhiPrasad:
Can't folks just use `beforeSend` to filter here? I'm not a huge fan of extending the public API this way, especially since then only electron will work this way.

timfish:
Ah yes, you're right!

PR #664, opened 2023-04-18
fix: IPC protocol should be registered as secure

### LINKED ISSUE #663: Sentry IPC protocol should be registered as secure

### How do you use Sentry?

Sentry Saas (sentry.io)

### Electron SDK Version

4.4.0

### Electron Version

21

Windows

### Link to Sentry event

_No response_

### Steps to Reproduce

Using Sentry Electron SDK with default options breaks page [Secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts).

Sentry IPC protocol should be registered as secure. `secure: true`
https://github.com/getsentry/sentry-electron/blob/master/src/main/ipc.ts#L134

### Expected Result

.

### Actual Result

.

timfish:
Closes #663

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -131,7 +131,7 @@ function configureProtocol(options: ElectronMainOptionsInternal): void {
   protocol.registerSchemesAsPrivileged([
     {
       scheme: PROTOCOL_SCHEME,
-      privileges: { bypassCSP: true, corsEnabled: true, supportFetchAPI: true },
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):
Can we add a regression test? Don't think we're gonna add more IPC, but better safe than sorry.

timfish:
To test this, I think we'd need to load the Electron renderer code from a secure context and to do that we'd need a valid SSL certificate.

AbhiPrasad:
Ah yeah that makes it tricky. Then I think it's fine, we just keep a close eye during reviews.

PR #710, opened 2023-07-29
feat: Tracing without performance

### LINKED ISSUE #688: Tracing without Performance
### Problem Statement
It should be possible that tracing information is propagated from/to incoming/outgoing HTTP requests even if performance is disabled and thus no transactions/spans are available.

### Solution Brainstorm

See exampling implementations in PHP, Java, and Python

To be done after JS: https://github.com/getsentry/sentry-javascript/issues/8352

timfish:
Closes #688

### DIFF
--- a/src/main/integrations/net-breadcrumbs.ts
+++ b/src/main/integrations/net-breadcrumbs.ts
@@ -1,7 +1,14 @@
 /* eslint-disable deprecation/deprecation */
+import { getDynamicSamplingContextFromClient } from '@sentry/core';
 import { getCurrentHub } from '@sentry/node';
 import { EventProcessor, Hub, Integration, Span, TracePropagationTargets } from '@sentry/types';
-import { fill, stringMatchesSomePattern } from '@sentry/utils';
+import {
+  dynamicSamplingContextToSentryBaggageHeader,
+  fill,
+  generateSentryTraceHeader,
+  logger,
+  stringMatchesSomePattern,
+} from '@sentry/utils';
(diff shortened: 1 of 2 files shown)


src/main/integrations/net-breadcrumbs.ts
```diff
@@ -202,6 +209,24 @@ function createWrappedRequestFactory(
           if (shouldAttachTraceData(method, url)) {
             request.setHeader('sentry-trace', span.toTraceparent());
```
AbhiPrasad, line 210: we need to add baggage here as well for this case, I guess we never ended up doing this.

timfish: oh yes, good spot!

test/unit/net.test.ts
```diff
@@ -217,4 +223,18 @@ describe.skip('net integration', () => {
...
+    it('adds headers without transaction', async () => {
+      createHubOnScope({
+        tracePropagationTargets: ['localhost'],
+        integrations: [new Net()],
+      });
+      const headers = await makeRequest();
+      const transaction = getActiveTransaction();
+
+      expect(transaction).to.be.undefined;
+      expect(headers['sentry-trace']).not.to.be.empty;
```
AbhiPrasad, line 237:
```suggestion
      expect(headers['sentry-trace']).not.to.be.empty;
      expect(headers['baggage']).not.to.be.empty;
```

AbhiPrasad (approved):

PR #715, opened 2023-08-03
feat: Optionally send sessions when they're created

### LINKED ISSUE #713: Sending initial session when created
In the JS SDK we send the session on creation (when initialized) https://github.com/getsentry/sentry-javascript/blob/develop/packages/browser/src/sdk.ts#L254C1-L255

This doesn't seem to be happening with Electron https://github.com/getsentry/sentry-electron/blob/85f82bc34baadb847ef0a1918dfcaf4b5f00449b/src/main/sessions.ts#L22

I think we need to send that initial session with `init: true` as per https://develop.sentry.dev/sdk/sessions/#important-client-behavior

@timfish this there some case I'm missing here?

timfish:
Closes #713

This PR adds an option to the `MainProcessSession` integration that forces sessions to be send immediatly on creation. Apart from a few lines to add this option and call `hub.captureSession()` when enabled, the rest of the changes are to add all the additional session envelopes to the session e2e tests.

In the Electron main process:
```ts
import { init, Integrations } from '@sentry/electron';

init({
  dsn: '__DSN__',
  integrations: [new Integrations.MainProcessSession({ sendOnCreate: true })]
});
```

### DIFF
--- a/src/main/integrations/main-process-session.ts
+++ b/src/main/integrations/main-process-session.ts
@@ -5,7 +5,11 @@ import { app } from 'electron';
 import { endSession, startSession } from '../sessions';
 
 interface Options {
-  /** Whether sessions should be sent immediately on creation */
+  /**
+   * Whether sessions should be sent immediately on creation
+   *
+   * @default false
+   */
   sendOnCreate?: boolean;
 }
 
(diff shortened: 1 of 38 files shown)


src/main/integrations/main-process-session.ts
```diff
@@ -4,6 +4,11 @@ import { app } from 'electron';
 
 import { endSession, startSession } from '../sessions';
 
+interface Options {
+  /** Whether sessions should be sent immediately on creation */
```
AbhiPrasad, line 8: Can we add that it defaults to false here in the doc string?

AbhiPrasad (approved):

PR #725, opened 2023-08-30
feat: Add integration to track sessions as app in foreground

### LINKED ISSUE #712: Stop sessions if app goes into background
### Problem Statement

https://develop.sentry.dev/sdk/sessions/

In the mobile SDKs, we end a session if the app is backgrounded for > 30s, and start another one if the app is foregrounded after that.

We might want to explore adding this functionality for Electron sessions.

### Solution Brainstorm

We can perhaps add this as an option to the main session integration and have it be opt-in. This way people's dashboards will not dramatically change when bumping SDK versions, and we can re-evaluate what the default should be when we bump the major version.

timfish:
Closes #712

This PR 
- Adds the `BrowserWindowSession` integration which tracks sessions as when the app has focus
- There is a timout to ensure that sessions are not ended immediatly if a user briefly switches away from the app
  - The default timeout is 30 seconds but this is configurable via the `backgroundTimeoutSeconds` constructor option.
- The session on exit handling was moved to `src/main/sessions.ts` since this is now use by both session integrations
- Some changes to the default integration logic were required to ensure only one session integration is used
- Errors or crashes while the app is in the background will not be attributed to any session

```ts
const { init, Integrations } = require('@sentry/electron');

init({
  dsn: '__DSN__',
  integrations: [new Integrations.BrowserWindowSession({ backgroundTimeoutSeconds: 10 })],
});
```


timfish:
I've limited this new feature to Electron >= v12 since older versions of Electron seem to unreliably report `focus` events which make this quite unreliable for sessions.

src/main/integrations/browser-window-session.ts
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * Number of seconds to wait before ending a session after the app loses focus.
+   *
+   * Default: 10 seconds
+   */
+  backgroundTimeoutSeconds?: number;
+}
+
+// The state can be, active, inactive, or waiting for a timeout
+type SessionState = 'active' | 'inactive' | { timer: NodeJS.Timeout };
```
AbhiPrasad, line 17: I prefer if this was a discriminated union so we don't have to have to typeof checks.

```ts
type SessionState =
  | {
      type: 'active';
    }
  | {
      type: 'inactive';
    }
  | {
      type: 'waiting';
      timer: NodeJS.Timeout;
    };
```

src/main/integrations/browser-window-session.ts
```diff
@@ -0,0 +1,97 @@
...
+      });
+    });
+
+    // if the app exits while the session is active, end the session
+    endSessionOnExit();
+  }
+
+  private _windowStateChanged = (): void => {
+    // We need to test all windows for visibility AND focus
+    const aWindowIsActive = BrowserWindow.getAllWindows().some((window) => window.isVisible() && window.isFocused());
```
AbhiPrasad, line 65: Can we not use `BrowserWindow.getFocusedWindow()` and check for visibility?

```suggestion
    const windowIsActive = BrowserWindow.getAllWindows().some((window) => window.isVisible() && window.isFocused());
```

AbhiPrasad (approved):
would be nice to get a test for errored session as well.

AbhiPrasad:
> Errors or crashes while the app is in the background will not be attributed to any session

The behaviour we kept here is good to go!

src/main/integrations/browser-window-session.ts
timfish: Yes good idea

src/main/integrations/browser-window-session.ts
timfish: I'll add a couple more tests and try it.

We want this to work with focus as well as visibility (hidden, minimised, etc)

I can't remember how or why but I came to this after quite a bit of testing.

AbhiPrasad (approved):

PR #738, opened 2023-09-11
feat: Update Sentry SDKs to v7.68.0

timfish:
As per the v7.68.0 changelog, this PR ensures that all auto instrumented events are now tagged with `mechanism.handled = false` which will impact crash-free sessions counts.


AbhiPrasad (approved):
We'll need to add an extra changelog note for this change specifically as well.

PR #743, opened 2023-09-12
fix: Limit `BrowserWindowSession` to Electron >= v14

timfish:
Release builds are tested on every Electron version and they seem to suggest that the current code for `BrowserWindowSession` does not correctly detect focus for Electron 12 and 13. This PR changes the minimum supported Electron version to v14.


AbhiPrasad:
Maybe it's worth vendoring in https://github.com/electron/electron/blob/650ef9a51cbaa3c65bea0428b897fe1dd5e5799e/lib/browser/api/browser-window.ts#L103-L110 because it seems there has been some bug fixes to the method

timfish:
Closing in favour of #744

Issue #729, opened 2023-08-31
Detect non-responsiveness (ANR) in Electron applications


cstavitsky:
### Problem Statement

Submitting on behalf of a customer:

> We have received feedback from our users regarding instances of unresponsiveness in our application. As such, we are keen on leveraging the capabilities of Sentry’s Electron SDK to implement ANR detection. 

The Electron SDK currently does not detect non-responsiveness/ANRs. 

#### Prior work: 

- Android ANR: https://docs.sentry.io/platforms/android/configuration/app-not-respond/
- Unity ANR: https://docs.sentry.io/platforms/unity/configuration/app-not-responding/
- App Hangs for iOS: https://docs.sentry.io/platforms/apple/configuration/app-hangs/

### Solution Brainstorm

> We are interested in understanding whether the SDK offers built-in capabilities to detect and handle application unresponsiveness.

An internal chat with Abhi on slack notes that we may be able to do watchdog approach that monitors the main thread but it’ll have to investigated more.
```[tasklist]
### Tasks
- [x] https://github.com/getsentry/sentry-javascript/pull/9079
- [x] https://github.com/getsentry/sentry-javascript/pull/9128
- [ ] https://github.com/getsentry/sentry-electron/pull/753
- [ ] https://github.com/getsentry/sentry-electron/pull/762
- [ ] https://github.com/getsentry/sentry-docs/pull/8146
- [ ] #770
- [ ] https://github.com/getsentry/sentry-javascript/pull/9268
```

AbhiPrasad:
@timfish could we do a quick investigation into how much time/effort this will take?

timfish:
Electron has the [`unresponsive` event](https://www.electronjs.org/docs/latest/api/web-contents#event-unresponsive ) for renderer processes.

By default [we currently only log breadcrumbs](https://github.com/getsentry/sentry-electron/blob/24bbeef183e2c1edda7248c21b1956d3727e819c/src/main/integrations/electron-breadcrumbs.ts#L69) for these events.

fredyangRC:
As we briefly discussed in the meeting with your CFO and Anthony, ANR (or app freeze) monitoring is as critical as crash monitoring. I see on your web site that other platforms (mobile, Mac) actually are supported but the Windows. This is an essential function we are looking for. 
We missed a major freeze issue in the recent release that caused a lot of customer complaints, and the analysis and troubleshooting was super hard due to lacking of this monitoring ability.

[...]

PR #753, opened 2023-09-28
feat: Add ANR detection for main process

timfish:
Ref #729

This PR adds ANR detection for the Electron main process:

```ts
import { init, enableAnrDetection } from '@sentry/electron';
 
init({ dsn: "__DSN__" });
 
// Electron v28+ with ESM
await enableAnrDetection({ mainProcess: { captureStackTrace: true }});
runApp();

// with CJS
enableAnrDetection({ mainProcess: { captureStackTrace: true }}).then(() => {
   runApp();
});
```

Should we keep the main/renderer config separate like this?


src/main/anr.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    options.mainProcess = options.mainProcess || {};
+
+    // We need to override the entryScript option to make it work with Electron which doesn't get passed a script in
+    // the process.argv when the app is packaged
+    if (options.mainProcess.entryScript === undefined) {
+      options.mainProcess.entryScript = app.getAppPath();
+    }
+
+    return enableNodeAnrDetection(options.mainProcess);
```
AbhiPrasad, line 52: let's avoid mutating the original options object. We removed this pattern in the js codebase.

AbhiPrasad (approved):
Looks great!

AlbertInRC:
Hi @tim @AbhiPrasad , will the ANR events impact the Crash Free Session rate?

timfish:
> will the ANR events impact the Crash Free Session rate?

No, not currently.

PR #755, opened 2023-09-29
test: Don't test all Electron versions for releases

timfish:
This PR changes it so we no longer test every version of Electron for release builds. v2 and the last 10 versions is more than enough.


timfish:
Closing this because I'm hoping #757 fixes it

Issue #751, opened 2023-09-25
SDK does not support ESM in main process


mahnunchik:
### Is there an existing issue for this?

- [X] I have checked for existing issues https://github.com/getsentry/sentry-javascript/issues
- [X] I have reviewed the documentation https://docs.sentry.io/
- [X] I am using the latest SDK release https://github.com/getsentry/sentry-javascript/releases

### How do you use Sentry?

Sentry Saas (sentry.io)

### Electron SDK Version

4.11.0

### Electron Version

28

### What platform are you using?

MacOS

### Steps to Reproduce

Electron 28 with ESM support

### Expected Result

Named export is supported as it shown in the readme

```js
import { init } from '@sentry/electron';

init({
  dsn: '__DSN__',
  // ...
});
```

### Actual Result

SyntaxError: Named export 'init' not found. The requested module '@sentry/electron' is a CommonJS module, which may not support all module.exports as named exports.

mahnunchik:
It seems `@sentry/electron` is completely broken when exported as ESM module in native ESM environment.

```js
import Sentry from '@sentry/electron';
```

```
node_modules/@sentry/electron/esm/main/index.js:1
import { ensureProcess } from '../common';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at internalCompileFunction (node:internal/vm:73:18)
    at wrapSafe (node:internal/modules/cjs/loader:1185:20)
    at Module._compile (node:internal/modules/cjs/loader:1227:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1326:10)
    at Module.load (node:internal/modules/cjs/loader:1126:32)
    at Module._load (node:internal/modules/cjs/loader:967:12)
    at f._load (node:electron/js2c/asar_bundle:2:13542)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:169:29)
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)
```

mahnunchik:
@timfish I can not use sentry with Electron 28 in ESM environment. It is impossible to import cjs version due to `exports` field in package.json

Could it be fixed?

timfish:
It was on the list to test but I wasn't aware of a release yet since v27 is still in beta.
https://github.com/getsentry/sentry-electron/issues/730

I see there's a nightly. Is that what you're using?

It should be relatively easy to fix and I need to add some integration tests too.

mahnunchik:
[...]

PR #756, opened 2023-09-29
feat: Node compatible ESM build

timfish:
This PR changes to using Rollup to build the output files.

This should fix main process ESM compatibility (#751) but since Electron v28 is still published under `electron-nightly` there is no easy way to add this to our test matrix until they publish a beta release.

I have tested this locally with a nightly build but I don't think it's worth the effort or adding nightly support to our e2e test suite when there will likely be a beta built in a few weeks.

- Because Rollup requires the TypeScript module target to be ESNext and then Rollup converts to CJS, this caused a few incompatibility issues with existing side effects.
  - TypeScript enums are evil but we can't change them until the next major
  - Some side-effects appear unavoidable to keep abnormal session tracking working correctly 
- Detection of code loaded in the incorrect process relied on require order which isn't compatible with ESM
- Now we're using Rollup we can bundle the preload scripts simply
- Preload injection required some changes to work without `require.resolve`


AbhiPrasad:
Gonna review this on Monday - having the bundled preload script is really nice.

timfish:
> having the bundled preload script is really nice

No more weird manual bundle creation!

timfish:
It's worth noting that this still isn't recognised as ESM compatible by Electron unless I put `{ "type": "module" }` in `./esm/package.json`!

src/main/integrations/preload-injection.ts
```diff
@@ -4,12 +4,34 @@ import { Integration } from '@sentry/types';
...
+  try {
+    return rendererRequiresCrashReporterStart()
+      ? require.resolve('../../preload/legacy.js')
+      : require.resolve('../../preload/index.js');
+  } catch (_) {
+    try {
+      // This could be ESM
+      const currentDir = fileURLToPath(import.meta.url);
+      // Use the CJS preload
+      return resolve(currentDir, '..', '..', '..', '..', 'preload', 'index.js');
```
AbhiPrasad, line 26: I wonder if this will break with any bundlers, but if the e2e tests pass we should be fine I guess.

src/main/electron-normalize.ts
```diff
@@ -20,13 +20,17 @@ export const isPackaged = (() => {
 
 /** A promise that is resolved when the app is ready */
 export const whenAppReady: Promise<void> = (() => {
-  return app.isReady()
-    ? Promise.resolve()
-    : new Promise<void>((resolve) => {
-        app.once('ready', () => {
-          resolve();
+  if (app) {
+    return app.isReady()
```
AbhiPrasad, line 24: when is app not defined?

AbhiPrasad (approved):

src/main/electron-normalize.ts
timfish: If this code gets loaded into the Electron renderer rather than the main process.

We have code that displays a warning if you load the wrong code into the wrong process. Rollup requires tha we target `module: esnext` and it then transpiles to CJS/ESM. The issue with this is that imports/side-effects are changed/moved and these bits of code get hit before our warning code!

PR #759, opened 2023-09-29
fix: Deprecate `EventToMain` integration

timfish:
This integration hasn't been used since we started routing events from renderer to main via a custom transport.

It was left behind because it is exported and removing it would technically be a breaking change. It should be marked as deprecated so nobody uses it!


AbhiPrasad (approved):
We should even think about removing it from the docs.

PR #761, opened 2023-09-30
fix: Re-work Sentry minidump loader

### LINKED ISSUE #763: Invalid Minidump

### How do you use Sentry?

Sentry Saas (sentry.io)

### Electron SDK Version

v4

### Electron Version

any

### Steps to Reproduce

There is a small stream of native crash event that are displayed as `Invalid minidump`:

image

Looking at the minidump attachments, these are all missing the minidump header. I could be that these events are being read before they have completed writing but it is unlikely since we should not get the event until this is complete.

timfish:
Closes #763

This PR re-writes the minidump loading code. 
- Now all the loading, validation, deleting and error catching logic is all contained in `createMinidumpLoader`. 
- Removes possible race conditions from overlapping calls to lookup minidumps via our existing `Mutex`
- Only reads minidumps from disk when they haven't been modified for 1 second
- Skips sending minidumps if they're too small or dont start with the expected minidump header
- Adds unit tests for the new logic

I had to extend the timings on many of the e2e tests because these changes add a delay in minidump sending while we wait to ensure the file is not being written.


test/unit/minidump-loader.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    setTimeout(() => {
+      expect(passedAttachment).to.be.false;
+      expect(existsSync(missingHeaderDump)).to.be.false;
+      expect(existsSync(tooSmallDump)).to.be.false;
+      done();
+    }, 2_000);
+  });
+
+  it("doesn't send minidumps that are over 30 days old", (done) => {
```
timfish, line 71: This test fails in CI on Windows and Linux but passes on macOS in CI and locally 🤷‍♂️

test/e2e/test-apps/sessions/native-crash-renderer/session-next.json
```diff
@@ -0,0 +1,17 @@
+{
```
timfish, line 1: I'm not 100% sure why this extra sessionm update gets sent.

It's another `ok` status with an `id` that matches the current session but with `init: false`.

src/main/integrations/sentry-minidump/minidump-loader.ts
```diff
@@ -1,73 +1,111 @@
...
+          logger.log('Found minidump', path);
 
-function createMinidumpLoader(fetchMinidumpsImpl: MinidumpLoader): MinidumpLoader {
-  const knownPaths: string[] = [];
+          let stats = await statAsync(path);
 
-  return async () => {
-    const minidumps = await fetchMinidumpsImpl();
-    logger.log(`Found ${minidumps.length} minidumps`);
+          const thirtyDaysAgo = new Date().getTime() - MAX_AGE_DAYS * 24 * 3_600 * 1_000;
```
AbhiPrasad, line 56: l: Let's lift magic number into global constant

test/unit/minidump-loader.test.ts
AbhiPrasad: yeah this is weird - so `utimesSync` doesn't work? 

Can we change file location for test?

test/e2e/test-apps/sessions/native-crash-renderer/session-next.json
AbhiPrasad: Is this from calling `checkPreviousSession` maybe?

src/main/integrations/sentry-minidump/minidump-loader.ts
```diff
@@ -1,73 +1,111 @@
...
-}
+/**
+ * Creates a minidump loader
+ * @param getMinidumpPaths A function that returns paths to minidumps
+ * @param preProcessFile A function that pre-processes the minidump file
+ * @returns A function to fetch minidumps
+ */
+export function createMinidumpLoader(
+  getMinidumpPaths: () => Promise<string[]>,
+  preProcessFile: (file: Buffer) => Buffer = (file) => file,
```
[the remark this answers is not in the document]
timfish: The crashpad loader uses the default (ie. pass through `(file) => file`) since only breakpad required pre-processing.

test/unit/minidump-loader.test.ts
timfish: Just tested Windows locally and that's not working either. Looking into it. 

Tried `Math.floor(new Date().getTime() / 1000)` in case it was the decimal that was upsetting it!

timfish:
My best guess is that the extra session has always been there but we're only seeing it now after extending the app run time to cater for the sending delays added by this PR

PR #762, opened 2023-10-02
feat: Identify `webContents` of renderers via custom protocol 

timfish:
The Electron SDK supports two ways for the renderers to send envelopes and scope to the main process.

- Electron IPC - Preferred but requires a preload script which we attempt to inject automatically but this fails when bundling the main process code
- Custom protocol + fetch - No setup required but Electron does not give us a way to determine which renderer the event came from (https://github.com/electron/electron/issues/31718)

This has meant custom protocol has not supported all the same context as Electron IPC because we're missing the reference to the `webContents`.

This is a problem for renderer ANR detection. We should be able to use the Electron [debugger](https://www.electronjs.org/docs/latest/api/debugger#debuggerattachprotocolversion) API to capture stack traces but once a renderer stops sending us mesages, we need to know which `webContents` to attach to! 

This PR adds a means for us to identify renderers over a custom protocol:
- In protocol mode, `window.__SENTRY_RENDERER_ID__` is set to a uuid
- When a renderer starts up and pings the main proces via the custom protocol, we use `executeJavaScript` to fetch the id and save the mapping between id and `webContents`
- When the renderer sends events/envelopes/scope and later ANR messages, the id is included in the headers
- In the main process, we can use the id in the headers to determine which `webContents` the message came from

### DIFF
--- a/src/common/ipc.ts
+++ b/src/common/ipc.ts
@@ -2,7 +2,7 @@ export const PROTOCOL_SCHEME = 'sentry-ipc';
 
 export enum IPCChannel {
   /** IPC to check main process is listening */
-  PING = 'sentry-electron.ping',
+  RENDERER_START = 'sentry-electron.renderer-start',
   /** IPC to send a captured event to Sentry. */
   EVENT = 'sentry-electron.event',
   /** IPC to pass scope changes to main process. */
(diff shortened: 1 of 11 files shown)


AbhiPrasad (approved):

timfish:
I want to add an extra test for this before merging

PR #787, opened 2023-11-30
fix: Patch `registerSchemesAsPrivileged` so sentry scheme isn't overwritten

### LINKED ISSUE #661: Sentry breaks custom protocols because it calls `registerSchemesAsPrivileged` which can be called only once

### How do you use Sentry?

Sentry Saas (sentry.io)

### Electron SDK Version

4.4.0

### Electron Version

24.0.0

MacOS

### Link to Sentry event

_No response_

### Steps to Reproduce

Create any custom protocol:

```ts
  protocol.registerBufferProtocol("foobar", (request, callback) => {
    return callback({ data: Buffer.from("Hello World") });
  });
```

Mark it as priviliged so it supports fetch API

```ts
protocol.registerSchemesAsPrivileged([
    {
      scheme: "foobar",
      privileges: { bypassCSP: true, corsEnabled: true, supportFetchAPI: true },
    },
  ]);
```

Initialize sentry main module (and renderer module)

```ts
import * as Sentry from "@sentry/electron/main";

export function initSentryMain() {
  console.info("Initializing Sentry");

  Sentry.init(sentryConfig);
}
```

### Expected Result

I am able to call my custom protocol with fetch API on renderer eg. `fetch("foobar://some-path");`

### Actual Result

After 2 totally ruined days of debugging this; having no clue it can be caused by sentry, I finally found that the issue happens because in `main/ipc.ts` of Sentry you do:

```ts
/** Enables Electron protocol handling */
function configureProtocol(options: ElectronMainOptionsInternal): void {
  if (app.isReady()) {
    throw new SentryError("Sentry SDK should be initialized before the Electron app 'ready' event is fired");
  }

  protocol.registerSchemesAsPrivileged([
    {
      scheme: PROTOCOL_SCHEME,
      privileges: { bypassCSP: true, corsEnabled: true, supportFetchAPI: true },
    },
  ]);
```

Which is incorrect. As Electron docs says https://www.electronjs.org/docs/latest/api/protocol#protocolregisterschemesasprivilegedcustomschemes 

> Note: This method can only be used before the ready event of the app module gets emitted **and can be called only once.**

As a result - you're silently overwriting previously defined schemes with your own, without warning. 

It results in 

```
Fetch API cannot load foobar://fooo. URL scheme "foobar" is not supported.
```

What makes it even more confusing is that in my case (and I imagine most of the cases) Sentry is only enabled in production, which made my custom protocols work just fine in dev.

Please fix it and please be careful when you use code that overwrites settings defined by the app. It caused a lot of stress and hair-pulling.

-------------------

## Workaround:

To workaround the issue and keep sentry working, AFTER initializing sentry, call

```ts
export function registerPriviligedProtocols() {
  protocol.registerSchemesAsPrivileged([
    {
      scheme: "sentry-ipc",
      privileges: { bypassCSP: true, corsEnabled: true, supportFetchAPI: true },
    },
    // <<--- REST OF MY CUSTOM SCHEMES HERE
  ]);
}
```

to register both sentry protocols and our own protocols.

timfish:
Closes #661

This fix patches the Electron `registerSchemesAsPrivileged` function so that subsequent calls by user code don't overwrite the Sentry scheme.

Unfortunately this Electron API only respects the last schemes passed before app ready and doesn't warn if called multiple times.

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -176,12 +176,13 @@ function configureProtocol(options: ElectronMainOptionsInternal): void {
 
   protocol.registerSchemesAsPrivileged([SENTRY_CUSTOM_SCHEME]);
 
-  // We override this function so that later user calls to registerSchemesAsPrivileged don't overwrite our custom scheme
+  // We Proxy this function so that later user calls to registerSchemesAsPrivileged don't overwrite our custom scheme
   // eslint-disable-next-line @typescript-eslint/unbound-method
-  const originalFn = protocol.registerSchemesAsPrivileged;
-  protocol.registerSchemesAsPrivileged = function (schemes) {
-    originalFn.call(this, [...schemes, SENTRY_CUSTOM_SCHEME]);
-  };
+  protocol.registerSchemesAsPrivileged = new Proxy(protocol.registerSchemesAsPrivileged, {
+    apply: (target, __, args: Parameters<typeof protocol.registerSchemesAsPrivileged>) => {
+      target([...args[0], SENTRY_CUSTOM_SCHEME]);
+    },
+  });
 
   const rendererStatusChanged = createRendererAnrStatusHandler();
 
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -12,12 +12,17 @@ import {
   PROTOCOL_SCHEME,
   RendererStatus,
 } from '../common';
-import { createRendererAnrStatusHook } from './anr';
+import { createRendererAnrStatusHandler } from './anr';
 import { registerProtocol, supportsFullProtocol, whenAppReady } from './electron-normalize';
 import { ElectronMainOptionsInternal } from './sdk';
 
 let KNOWN_RENDERERS: Set<number> | undefined;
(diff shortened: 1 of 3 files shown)


src/main/ipc.ts
```diff
@@ -169,14 +174,16 @@ function configureProtocol(options: ElectronMainOptionsInternal): void {
...
-      privileges: { bypassCSP: true, corsEnabled: true, supportFetchAPI: true, secure: true },
-    },
-  ]);
+  protocol.registerSchemesAsPrivileged([SENTRY_CUSTOM_SCHEME]);
+
+  // We override this function so that later user calls to registerSchemesAsPrivileged don't overwrite our custom scheme
+  // eslint-disable-next-line @typescript-eslint/unbound-method
+  const originalFn = protocol.registerSchemesAsPrivileged;
+  protocol.registerSchemesAsPrivileged = function (schemes) {
+    originalFn.call(this, [...schemes, SENTRY_CUSTOM_SCHEME]);
```
AbhiPrasad, line 183: let's use `new Proxy` instead?

PR #797, opened 2023-12-04
fix: Don't `app.getPath('userData')` before SDK `init`

### LINKED ISSUE #796: `app.getPath('userData')` is called before user calls `init`

### How do you use Sentry?

Sentry Saas (sentry.io)

### Electron SDK Version

4.15.0

### Electron Version

All

All

### Steps to Reproduce

https://github.com/getsentry/sentry-electron/blob/aca8dd4645e85037472e8ff4d3f3e0b0fef91760/src/main/fs.ts#L5-L7

### Expected Result

This means that if users call `app.setPath('userData', path)` to set a custom user directory, Sentry will continue using the previous directory.

### Actual Result

`app.getPath` should not be called until the SDKs `init` is called

timfish:
Fixes #796

### DIFF
--- a/src/main/fs.ts
+++ b/src/main/fs.ts
@@ -3,7 +3,10 @@ import { mkdir, readdir, readFile, stat, statSync, unlink, writeFile } from 'fs'
 import { dirname, join, resolve } from 'path';
 import { promisify } from 'util';
 
-export const sentryCachePath = join(app ? app.getPath('userData') : '', 'sentry');
+/** Gets the Sentry Cache path */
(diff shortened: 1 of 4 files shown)


AbhiPrasad (commented):
Can we add a test that validates this?

AbhiPrasad (approved):

lforst (approved):

timfish:
The new test passes on my Windows machine but fails in CI 🤦🏻‍♂️

PR #799, opened 2023-12-11
feat: Browser profiling

### LINKED ISSUE #792: Browser Profiling
### Problem Statement

Newer versions of Chrome support the `BrowserProfilingIntegration` which captures CPU profiles. 

### Solution Brainstorm

Initial testing shows this does not work out-of-the-box.

timfish:
Closes #792

Adds a new `RendererProfiling` integration:
- Requires Electron >= v15 (For Chrome 94)
- Enabled via the new `enableRendererProfiling` option
- It adds `'Document-Policy': 'js-profiling'` headers to all renderer requests
- Caches profile envelope items as they arrive in the main process from renderers
- Hooks `beforeEnvelope` to re-add the profiles to the correct envelopes

This PR also adds a test to ensure a profile is sent with the transaction.

## Enabling Browser Profiling

From a users perspective, to enable profiling in the renderers, you'll need to pass `enableRendererProfiling: true` in the main process:

```ts
const { init } = require('@sentry/electron/main');

init({
  dsn: '__DSN__',
  enableRendererProfiling: true,
});
```

Then in the renderer follow the regular browser profiling instructions:
```ts
import { addTracingExtensions, init, BrowserProfilingIntegration } from '@sentry/electron/renderer';

addTracingExtensions();

init({
  integrations: [new BrowserProfilingIntegration()],
  tracesSampleRate: 1,
  profilesSampleRate: 1,
});
```

### DIFF
--- a/src/main/integrations/renderer-profiling.ts
+++ b/src/main/integrations/renderer-profiling.ts
@@ -1,0 +1,120 @@
+import { NodeClient } from '@sentry/node';
+import { Event, Integration, Profile } from '@sentry/types';
+import { forEachEnvelopeItem, LRUMap } from '@sentry/utils';
+import { app } from 'electron';
+
+import { normaliseProfile } from '../../common';
+import { getDefaultEnvironment, getDefaultReleaseName } from '../context';
+import { ELECTRON_MAJOR_VERSION } from '../electron-normalize';
+import { ElectronMainOptionsInternal } from '../sdk';
+
+// A cache of renderer profiles which need attaching to events
+let RENDERER_PROFILES: LRUMap<string, Profile> | undefined;
+
+/**
+ * Caches a profile to later be re-attached to an event
+ */
+export function rendererProfileFromIpc(event: Event, profile: Profile): void {
+  if (!RENDERER_PROFILES) {
+    return;
+  }
+
+  const profile_id = profile.event_id;
+  RENDERER_PROFILES.set(profile_id, profile);
+
+  if (event) {
+    event.contexts = {
+      ...event.contexts,
+      // Re-add the profile context which we can later use to find the correct profile
+      profile: {
+        profile_id,
+      },
+    };
+  }
+}
+
+/**
+ * Injects 'js-profiling' document policy headers and ensures that profiles get forwarded with transactions
+ */
+export class RendererProfiling implements Integration {
+  /** @inheritDoc */
+  public static id: string = 'RendererProfiling';
+
+  /** @inheritDoc */
+  public readonly name: string;
+
+  public constructor() {
+    this.name = RendererProfiling.id;
+  }
+
+  /** @inheritDoc */
+  public setupOnce(): void {
+    //
+  }
+
+  /** @inheritDoc */
+  public setup(client: NodeClient): void {
+    const options = client.getOptions() as ElectronMainOptionsInternal;
+    if (!options.enableRendererProfiling) {
+      return;
+    }
+
+    if (ELECTRON_MAJOR_VERSION < 15) {
+      throw new Error('Renderer profiling requires Electron 15+ (Chromium 94+)');
+    }
+
+    RENDERER_PROFILES = new LRUMap(10);
+
+    app.on('ready', () => {
+      // Ensure the correct headers are set to enable the browser profiler
+      for (const sesh of options.getSessions()) {
+        sesh.webRequest.onHeadersReceived((details, callback) => {
+          callback({
+            responseHeaders: {
+              ...details.responseHeaders,
+              'Document-Policy': 'js-profiling',
+            },
+          });
+        });
+      }
+    });
+
+    // Copy the profiles back into the event envelopes
+    client.on('beforeEnvelope', (envelope) => {
+      let profile_id: string | undefined;
+
+      forEachEnvelopeItem(envelope, (item, type) => {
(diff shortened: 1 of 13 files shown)


src/main/integrations/renderer-profiling.ts
```diff
@@ -0,0 +1,120 @@
...
+      if (!profile) {
+        return;
+      }
+
+      normaliseProfile(profile, app.getAppPath());
+      profile.release = options.release || getDefaultReleaseName();
+      profile.environment = options.environment || getDefaultEnvironment();
+
+      // @ts-expect-error untyped envelope
+      envelope[1].push([{ type: 'profile' }, profile]);
```
AbhiPrasad, line 117: we should type this envelope, could we upstream change?

src/main/integrations/renderer-profiling.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  /** @inheritDoc */
+  public setupOnce(): void {
+    //
+  }
+
+  /** @inheritDoc */
+  public setup(client: NodeClient): void {
+    const options = client.getOptions() as ElectronMainOptionsInternal;
+    if (!options.enableRendererProfiling) {
```
AbhiPrasad, line 58: Can we instead gate this by `profilesSampleRate` being defined?

test/e2e/test-apps/other/browser-profiling/src/index.html
```diff
@@ -0,0 +1,47 @@
...
+  <head>
+    <meta charset="UTF-8" />
+  </head>
+  <body>
+    <script>
+      const crypto = require('crypto');
+      const {
+        addTracingExtensions,
+        init,
+        BrowserProfilingIntegration,
```
AbhiPrasad, line 12: We need both `RendererProfiling` and `BrowserProfilingIntegration`?

AbhiPrasad (commented):
awesome

src/main/integrations/renderer-profiling.ts
timfish: There are 2 processes here. Using the sample rate would enable profiling in the main process.

test/e2e/test-apps/other/browser-profiling/src/index.html
timfish: `RendererProfiling` is only for the main node process and enables the header injection.

timfish:
I added some more detail to the top post to show how this would be used by users

src/main/integrations/renderer-profiling.ts
timfish: A PR for this:
https://github.com/getsentry/sentry-javascript/pull/9798

AbhiPrasad: We released the types changes in https://github.com/getsentry/sentry-javascript/releases/tag/7.87.0

timfish: Updating to the latest JavaScript SDKs is blocked on scope changes. Shall we merge this PR as it is?
https://github.com/getsentry/sentry-javascript/pull/9801#issuecomment-1855560528

src/main/integrations/renderer-profiling.ts
JonasBa: @timfish Instead of requiring users to both enable renderer profiling bool and BrowserProfiling integration, can we automatically inject BrowserProfiling if we can infer enableRendererProfiling bool is set to true? I'm not sure if this is possible in the concept of electron and the rendering model, but it would allow us to reduce the config down to a single configuration point.

JonasBa:
No blocking comments, just wondering if we can reduce the config point down to a single entry

timfish:
> can we automatically inject BrowserProfiling 

Electron has a "preload" context where you can inject code but it's 
- isolated from the browser context
- will only load from a path which isn't compatible with bundling.

We can inject code into the browser context via either the debugger interface or via intercepting the requests for existing assets (html, js, etc).

The debugger is a bit heavy handed and likely won't happen before `init` has occurred.

Assuming the browser front end is loading from a `file://` URI, we could attempt to intercept and inject js code but which files should we inject into? Can we inject a script into all html files? 

If we could determine a reliable way to do this, we could inject all the browser code automatically, not just additional features like this!

Due to (optional but common) bundling of the main node process code, any code we inject would need to be inlined into the main process SDK source and we'd need add this in a way that doesn't impact app size for those not using this feature.

JonasBa:
I think it's very reasonable to ask people to do two different configuration changes in this context for now, the isolation seems unreliable to work around, but could be a DX improvement if we can do it reliably.

JonasBa (approved):

AbhiPrasad (approved):

src/main/integrations/renderer-profiling.ts
```diff
@@ -0,0 +1,120 @@
...
+    RENDERER_PROFILES = new LRUMap(10);
+
+    app.on('ready', () => {
+      // Ensure the correct headers are set to enable the browser profiler
+      for (const sesh of options.getSessions()) {
+        sesh.webRequest.onHeadersReceived((details, callback) => {
+          callback({
+            responseHeaders: {
+              ...details.responseHeaders,
+              'Document-Policy': 'js-profiling',
```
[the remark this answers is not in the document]
timfish: Ah good point!

PR #800, opened 2023-12-12
feat: Update JavaScript SDKs to v7.90.0

timfish:
This PR updates to v7.90.0 of the JavaScript SDKs. The update to v7.91.0 will come in another PR since this requires changes for ANR and I'd like to keep that seperate.

- Migrates to storing the new `ScopeData` rather than using private `Scope` properties
- Fixes all the scope API deprecations
- Fixes all the new lint failures
- Ensures we export all the new APIs!


src/main/integrations/additional-context.ts
```diff
@@ -44,21 +44,26 @@ export class AdditionalContext implements Integration {
     addGlobalEventProcessor(async (event: Event) => this._addAdditionalContext(event));
 
     // Some metrics are only available after app ready so we lazily load them
-    void whenAppReady.then(() => {
-      const { language, screen } = this._options;
+    whenAppReady.then(
```
lforst, line 47: Just fyi, when you're 100% certain that a promise cannot reject it is also fine to just ignore the lint rule!

lforst (approved):

PR #808, opened 2024-01-03
feat: Migrate to functional integration APIs

timfish:
Migrates all the integrations to use the new functional style.

### DIFF
--- a/src/main/integrations/electron-breadcrumbs.ts
+++ b/src/main/integrations/electron-breadcrumbs.ts
@@ -1,6 +1,6 @@
-import { addBreadcrumb, getCurrentHub } from '@sentry/core';
+import { addBreadcrumb, convertIntegrationFnToClass } from '@sentry/core';
 import { NodeClient } from '@sentry/node';
-import { Breadcrumb, Integration } from '@sentry/types';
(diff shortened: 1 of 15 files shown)


src/main/integrations/electron-breadcrumbs.ts
```diff
@@ -202,4 +142,57 @@ export class ElectronBreadcrumbs implements Integration {
...
+
+      whenAppReady.then(
+        () => {
+          // We can't access these until app 'ready'
+          if (options.screen) {
+            patchEventEmitter(screen, 'screen', options.screen);
+          }
+
+          if (options.powerMonitor) {
+            patchEventEmitter(powerMonitor, 'powerMonitor', options.powerMonitor);
```
mydea, line 161: just to be sure, I don't really know how exactly that works for Electron, but is it a problem if we patch this multiple times? Can there even be multiple clients in electron?

mydea: 👍

mydea (approved):
Just skimmed this TBH but overall looks good to me!

timfish:
We've got quite good e2e test coverage. I broke a couple in the process of this PR!

PR #811, opened 2024-01-05
feat: Add support for performance metrics

### LINKED ISSUE #803: Add support for Sentry Developer Metrics
### Problem Statement

Sentry is introducing support for a [metrics product](https://github.com/getsentry/sentry/discussions/58584).

Spec: https://develop.sentry.dev/sdk/metrics/

This was already implemented in the JavaScript SDK:
- Browser: https://github.com/getsentry/sentry-javascript/discussions/9938
- Node: https://github.com/getsentry/sentry-javascript/discussions/9973

Min required JS SDK version: `7.91.0`

### Solution Brainstorm

Let's add this to the Electron SDK!

This should be similar to the Node SDK, so we gate the usage by `metricsAggregator` experiment.

```ts
Sentry.init({
  dsn: '__DSN__',
  _experiments: {
    metricsAggregator: true,
  },
});
```

I think the Electron Client can just use the server-runtime metrics aggregator, the renderer can just forward to the main process, and the main process client is in charge of the aggregation.

timfish:
Closes #803

To enable metrics, you follow the same instructions as for node in the main process:
```ts

Sentry.init({
  dsn: '__DSN__',
  _experiments: {
    metricsAggregator: true,
  },
});

Sentry.metrics.increment('hits', 6);
```

In the renderer, no metric specific config is required:
```ts
import * as Sentry from '@sentry/electron/renderer';

Sentry.init();

Sentry.metrics.increment('hits', 6);
```


AbhiPrasad:
Let's avoid having the user configure a `MetricsAggregator` for the renderer. We did this for treeshaking purposes in browser, but I think we can get away from that here.

The renderer can just configure it's metrics aggregator on init.

timfish:
> The renderer can just configure it's metrics aggregator on init.

Currently the main/renderer inits just call their respective node/browser inits. There are no derived clients.

It's probably more flexible to just add the `MetricsAggregator` integration to the default renderer integrations. This would also mean that if users copy the renderer instructions and use `new Sentry.metrics.MetricsAggregator()`, one integration will get dropped and they don't end up using the browser implementation.

AbhiPrasad (approved):

PR #839, opened 2024-02-29
feat: Support and test `Sentry.sendFeedback`

### LINKED ISSUE #834: Test User Feedback with Electron SDK
### Problem Statement

Sentry has user feedback functionality in alpha! https://docs.sentry.io/platforms/javascript/user-feedback/

### Solution Brainstorm

Let's test it out to see if it works with Electron.

If it works fine we can update in-app onboarding and docs to add electron support!

No rush on this task - lower prio compared to v8 stuff, but give it a try whenever you get cycles.

timfish:
Closes #834

This PR ensures that feedback can be sent from the renderer process and adds a test to ensure the envelopes correctly get passed through the main process.

### DIFF
--- a/src/common/normalize.ts
+++ b/src/common/normalize.ts
@@ -34,6 +34,10 @@ export function normalizeEvent(event: Event, basePath: string): Event {
     request.url = normalizeUrlToBase(request.url, basePath);
   }
 
+  if (event.contexts?.feedback?.url && typeof event.contexts.feedback.url === 'string') {
+    event.contexts.feedback.url = normalizeUrlToBase(event.contexts.feedback.url, basePath);
+  }
+
   event.contexts = {
     ...event.contexts,
     runtime: {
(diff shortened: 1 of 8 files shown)


AbhiPrasad (approved):

billyvg (approved):
I'm pretty sure that the backend will ignore any additional contexts, but will double check for you.

billyvg:
Also as a headsup, we're looking at consolidating `captureUserFeedback` with `sendFeedback` (and likely killing the latter API) for v8

timfish:
> we're looking at consolidating `captureUserFeedback` with `sendFeedback`

Yep no problem, will update all these tests when it happens. 

I think the main difference (in features/behaviour at least!) is that `sendFeedback` doesn't have a way to attach a related event id?

billyvg:
> I think the main difference (in features/behaviour at least!) is that `sendFeedback` doesn't have a way to attach a related event id?

Yep exactly

PR #846, opened 2024-03-15
feat: Limit minidumps sent per app session

timfish:
Since the Dedupe integration cannot block sending of duplicate minidumps, there is effectively no limit to the number of duplicate minidumps that can be sent per session. 

This PR add a `maxMinidumpsPerSession` option to the integration that defaults to 10.

Internal doc: https://www.notion.so/sentry/096c7ce2987a4e598ca234c82a222853?v=693bea0c95b54382bb9523ab05711127&p=1a1abe81a8ca4689b831efe8bbbb87d9&pm=s


AbhiPrasad (approved):
is this easy to test?

PR #863, opened 2024-03-27
feat: Check package exports (v5)

timfish:
This PR:
- Adds a script to ensure that both the main and renderer exports include the same exports as the node and browser SDKs respectively
- Adds missing exports for both main and renderer
- Removes all the noop `setupOnce` which I missed form a previous PR
- Removes the legacy preload script which is no longer required


scripts/check-exports.mjs
```diff
@@ -0,0 +1,50 @@
...
+
+const browserExports = Object.keys(browser);
+const rendererExports = Object.keys(renderer);
+const nodeExports = Object.keys(node);
+const mainExports = Object.keys(main);
+
+const ignoredBrowser = [
+  'SDK_VERSION',
+  'WINDOW',
+  'Integrations',
```
AbhiPrasad, line 20: We should mark a todo to remove this, it will be removed in the beta!

scripts/electron-shim.mjs
```diff
@@ -0,0 +1,58 @@
...
+      };
+    }
+
+    return nextResolve(specifier);
+  }
+  `);
+
+process.versions.electron = '29.1.0';
+
+register(new URL(`data:application/javascript;base64,${hookScript.toString('base64')}`), import.meta.url);
```
AbhiPrasad, line 58: this is pretty sweet that it works

@timfish you should do a little writeup and put it on https://github.com/getsentry/sentry.engineering!

AbhiPrasad (approved):

PR #866, opened 2024-04-03
feat: Migrate to Node context

### LINKED ISSUE #525: Add support for Culture Context
https://develop.sentry.dev/sdk/event-payloads/contexts/#culture-context

The [language field in device context was deprecated](https://develop.sentry.dev/sdk/event-payloads/contexts/#device-context),

### LINKED ISSUE #603: `user.ip_address` should not be set to `{{auto}}`
https://develop.sentry.dev/sdk/data-handling/#sensitive-data says:

> SDKs don't set `{{auto}}` as `user.ip`. This instructs the server to keep the connection's IP address.*

Currently the Electron SDK does this:
https://github.com/getsentry/sentry-electron/blob/c22dd572c8fb955e775de2cecd191c67d689623d/src/main/context.ts#L304

timfish:
Closes #525 
Closes #603

This PR migrates to using device context mostly supplied by the `nodeContextIntegration` rather than the similar implementation that it was originally forked from in the Electron SDK.

This PR:
- Moves `language` to `culture` context (#525)
- Removes `user.ip_address: '{{auto}}'` which should not be included by default (#603)
- Removes the `event_type` tag which was replaced with `event.environment` in v4
- The `mainContextIntegration` has been removed and replaced by:
  - `electronContextIntegration` which both adds Electron context and removes Node context which is not appropriate 
  - `normalizePathsIntegration` which normalises all paths 
- All context is now applied to events through event processors


AbhiPrasad (approved):

timfish:
For some reason the tests for v15.5.7 on ubuntu pass but the process hangs at the end and doesn't exit 😭

AbhiPrasad:
Stale promise somewhere?

might be useful to throw a https://www.npmjs.com/package/why-is-node-running at it and see what gets logged out.

Side note - can we max times for the tests? Ideally we don't use 5 hours for the runner, we can just cancel the tests after 30 min.

timfish:
> can we max times for the tests? 

Yeah it [looks like you can set a timeout for actions steps](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes) and 25m should be plenty.

timfish:
I have no idea how I managed to merge this. I didn't click anything and all the tests weren't passing

PR #870, opened 2024-04-03
fix: Failing CI tests

timfish:


AbhiPrasad (approved):

timfish:
I've forced mocha to `--exit` for now. 

I will revisit this if it's still an issue when I replace mocha with vitest.

PR #873, opened 2024-04-06
feat: Remove fs compatibility wrappers (v5)

timfish:
Now the supported Node versions have `fs` that exports `promises` and `mkdir` that supports `{ recursive: true }`, we no longer need the compatibility wrappers to support older Node versions.


lforst (approved):
Satisfying

PR #879, opened 2024-04-13
feat: Use `NodeClient` rather than `init` (v5)

timfish:
The Node `init` uses the Node `getDefaultIntegrations` which contains the auto performance integrations which are almost never relevant to Electron. This impacts tree shaking and will result in quite a large minimum bundle size for the Electron main process.

By using the `NodeClient` directly we can define more reasonable defaults for Electron.

This PR currently has to add `@sentry/opentelemetry` as a direct dependency to access `setOpenTelemetryContextAsyncContextStrategy`. Once https://github.com/getsentry/sentry-javascript/pull/11585 is ~~merged~~ released, this will not be required.


src/main/integrations/screenshots.ts
```diff
@@ -2,21 +2,13 @@ import { defineIntegration } from '@sentry/core';
...
-
 /**
  * Captures and attaches screenshots to events
  */
 export const screenshotsIntegration = defineIntegration(() => {
   return {
     name: 'Screenshots',
-    async processEvent(event, hint, client) {
-      const attachScreenshot = !!(client.getOptions() as ElectronMainOptions).attachScreenshot;
-
```
AbhiPrasad, line 15: I think we should still gate this here behind options in case people dynamically add integrations.

AbhiPrasad (approved):

src/main/integrations/screenshots.ts
timfish: Ok, will add these back!

PR #905, opened 2024-05-12
feat: Use yarn v4

timfish:
To enable yarn v4 and have it work with Volta I had to:

```
npm install -g corepack

corepack enable --install-directory ~/.volta/bin

yarn set version berry

yarn install
```


timfish:
For some reason the rollup build hangs at the end occasionally on macOS. I've seen this locally but can no longer reproduce it. It's strange because when it fails in a runner (and it doesn't fail in all of them), re-running always fails to complete.

Issue #931, opened 2024-06-27
Anomaly in reported crashed sessions


jjbayer:
A user reported a negative crash-free rate. We figured out that there's a single 10-second interval in which a very high number of crashed sessions was reported, but the number of _total_ sessions remained stable:

![image](https://github.com/getsentry/sentry-electron/assets/1565449/61d66f42-c27d-45e3-941f-0c9fbc1d878f)

This looks suspicious for two reasons:

1. The number itself seems unreasonably high.
2. `sessions.all` must always be larger than `sessions.crashed`. If the crashed number is higher, it means that the SDK reported session updates with `init: false` for which it has never reported the initial update with `init: true`.

Could there be a reporting bug in the electron SDK which causes large numbers of session updates to be flushed without `init: true`?

Zendesk ticket with URL to the API request shown above: https://sentry.zendesk.com/agent/tickets/124592

AbhiPrasad:
Closing this for clean up - please re-open if the issue is still there!

Issue #895, opened 2024-04-26
Sudden spike in the number of sessions per user in newer version of electron sdk


gid-sentry:
### Is there an existing issue for this?

- [X] I have checked for existing issues https://github.com/getsentry/sentry-javascript/issues
- [X] I have reviewed the documentation https://docs.sentry.io/
- [X] I am using the latest SDK release https://github.com/getsentry/sentry-javascript/releases

### How do you use Sentry?

Sentry Saas (sentry.io)

### Electron SDK Version

4.22.0

### Electron Version

27.2.0

### What platform are you using?

None

### Link to Sentry event

https://jupiter-ct.sentry.io/releases/desktop%4024.2.15%2BRCAppDesktop/?chart=sessionCount&issuesType=unhandled&pageEnd=2024-04-19T11%3A00%3A00&pageStart=2024-04-14T23%3A29%3A35&project=1430058

### Steps to Reproduce

Update to sdk version 4.22.0, have max Minidumps per session set to 8.

### Expected Result

Same rough number of sessions to users

### Actual Result

Significantly more sessions per user

gid-sentry:
Does not look like version 4.24 has updates that would affect this functionality as it seems like dependency updates for the javascript sdk + some replay related fixes.

gid-sentry:
Internal Case Doc: https://www.notion.so/sentry/096c7ce2987a4e598ca234c82a222853?v=693bea0c95b54382bb9523ab05711127&p=98b0d84eada9433fa8d313b613e9324c&pm=s

AlbertInRC:
Maybe you could check how the sessions were uploaded from a minimum group of users.

smeubank:
hi @timfish fyi the init looks something like this, would you mind having a look if anything is suspicious here, I am not familiar with this feature, or why there are 2 variants of minidump 😅 

```
import * as Sentry from "@sentry/electron";

Sentry.init({
  dsn: SENTRY_DSN,
  ….
  integrations: [
    new Sentry.Integrations.MainProcessSession({ sendOnCreate: true }),
    new Sentry.Integrations.ElectronMinidump()
  ],
 …..
});
```

timfish:
That init code looks fine. The `ElectronMinidump` integrations uses Electrons built in native uploader. It's potentially more reliable but lacks the ability to include full context like our custom uploader used in the `SentryMinidump` integration. 

Above, both 4.24 and 4.22.0 are mentioned. Which version is has caused the issue?

[...]

PR #961, opened 2024-08-17
fix: Duplicate crashed sessions

### LINKED ISSUE #941: The crash count doesn't match

### How do you use Sentry?

Sentry Saas (sentry.io)

### Electron SDK Version

4.22.0

### Electron Version

27.2.20

None

### Link to Sentry event

https://jupiter-ct.sentry.io/discover/homepage/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&project=1430058&query=release%3Adesktop%4024.2.360%2BRCAppDesktop+error.mechanism%3Aminidump&queryDataset=discover&sort=-timestamp&statsPeriod=90d&yAxis=count%28%29

### Steps to Reproduce

1. from the release page, there are 100 crash events.
image

2. From the discover page, there are 151 crash events.
image

### Expected Result

These two data should be equal.

### Actual Result

These two data are not equal.

timfish:
Closes #941. Maybe also #931 and #895.

I re-read through all the Sentry SDK development docs around sessions and found [this](https://develop.sentry.dev/sdk/sessions/#terminal-session-states):

> A session can exist in two states: in progress or terminated. A terminated session must not receive further updates. `exited`, `crashed` and `abnormal` are all terminal states. When a session reaches this state the client must not report any more session updates or start a new session.

The Electron SDK was not following this as it was sending further updates after the initial crashed session was sent. I would not have expected these updates to be counted as new sessions since the `sid` doesn't change and `init` is never `true` more than once. 

Make no assumptions and read the docs thoroughly!


Lms24 (approved):
Good catch! I had no idea about this either but TIL

PR #991, opened 2024-10-01
feat: Support Electron utility process

### LINKED ISSUE #987: Support Electron utility processes
### Problem Statement

electron apps can have utility processes: https://github.com/electron/electron/blob/main/docs/tutorial/process-model.md#the-utility-process

it's not clear whether importing `@sentry/electron/main` or `@sentry/electron/renderer` is correct

### Solution Brainstorm

a new `@sentry/electron/utility` would be ideal from a user perspective

perhaps for now using `@sentry/node` makes the most sense?

timfish:
Closes #987

`utility.js`
```ts
import * as Sentry from '@sentry/electron/utility';

// No config required in utility process
Sentry.init();
```

`main.js`
```ts
import * as path from 'path';
import * as Sentry from '@sentry/main';
import { app, utilityProcess } from 'electron';

Sentry.init({
  dsn: '__DSN__'
});

app.on('ready', () => {
  // No special config required in the main process, just fork as before
  const child = utilityProcess.fork(path.join(__dirname, 'utility.js'));
});
```

## How it works

We wrap `electron.utilityProcess.fork` which allows us to set up a message port between the processes. All events from the utility process are passed to the main process via a custom transport. This way we retain the offline support, events get the full Electron context and all events go through the main process `beforeSend` callback.

Unlike renderer processes, utility processes do not yet sync scope (ie. user, tags, extra, etc) to the main process.


AbhiPrasad (approved):
Let's make sure to update docs. Should we update https://github.com/getsentry/sentry-wizard as well?

timfish:
Will definitely need adding to the docs. Not sure this Electron feature is widely used enough to warrant adding to the wizard 🤔

PR #1024, opened 2024-11-29
fix: Sync global scope to main process

### LINKED ISSUE #1022: Sync global scope from render process to main process
### Problem Statement

It seems only current scope and isolated scope are synced to main process. Why not to sync global scope to main process?
[https://github.com/getsentry/sentry-electron/blob/master/src/common/scope.ts](url)

### Solution Brainstorm

Can we add `getGlobalScope().addScopeListener`? Or is there any problem to do so?

timfish:
Closes #1022 

To PR reviewers... am I merging the scopes with the correct hierarchy?


timfish:
Perfect, thanks for the link!

PR #1046, opened 2025-01-07
feat: Update JavaScript SDKs to v8.48.0

### LINKED ISSUE #1011: ANR Events not Symbolicating

- [x] I have checked for existing issues https://github.com/getsentry/sentry-javascript/issues
- [x] I have reviewed the documentation https://docs.sentry.io/
- [x] I am using the latest SDK release https://github.com/getsentry/sentry-javascript/releases

### How do you use Sentry?

Sentry Saas (sentry.io)

### Electron SDK Version

5.6.0

### Electron Version

32.1.2

Windows

### Link to Sentry event

https://openai.sentry.io/issues/6020361972/?environment=development&project=4507771614265344&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=30m&stream_index=1

### Steps to Reproduce

1. Build typescript with vite framework and use sentryVitePlugin configured as so:
```
sentryVitePlugin({
      org: "Redacted",
      project: "Redacted",
      authToken: process.env["SENTRY_AUTH_TOKEN"],
      telemetry: false,
      release: {
        name: pkg.version,
        inject: false,
      },
      sourcemaps: {
        filesToDeleteAfterUpload: [".vite/build/*.map"],
      },
    }),
```
2. Upload Native Symbols via a speerate script configured in package.json as:
`"symbol-upload": "sentry-cli debug-files upload -o Redacted -p Redacted packages"`

3. Simulate ANR issue in Electron-based application and send event to Sentry

### Expected Result

Error's callstack should symbolicate. Debug Symbols mapping to the issue are present in its release, verified built via Source Map Upload Report.

### Actual Result

ALL ANRs are unsymbolicated and claims that stack frame doesn't have Debug IDs when expanding the "unminify code" buttons for any stack frame.

Non-ANR events which are symbolicated contain a debug_meta/images key with the debug ids of the sourcemaps needed to symbolicate the event. By contrast, ANRs simply can't reference the images, even though we've verified they're there.

timfish:
Updates JavaScript SDKs to v8.48.0.

The Node SDK has a fix for ANR debug IDs so this PR also includes test changes to confirm this.

- Closes #1011

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,7 @@ jobs:
       # we want that the matrix keeps running, default is to cancel all if one fails.
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-22.04, macos-latest, windows-latest]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
@@ -73,7 +73,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-22.04, windows-latest, macos-latest]
         electron: ${{ fromJson(needs.build.outputs.matrix) }}
     env:
       ELECTRON_VERSION: ${{ matrix.electron }}
--- a/package.json
+++ b/package.json
@@ -60,16 +60,16 @@
     "e2e": "xvfb-maybe vitest run --root=./test/e2e --silent=false --disable-console-intercept"
   },
   "dependencies": {
-    "@sentry/browser": "8.47.0",
-    "@sentry/core": "8.47.0",
-    "@sentry/node": "8.47.0",
+    "@sentry/browser": "8.48.0",
+    "@sentry/core": "8.48.0",
+    "@sentry/node": "8.48.0",
     "deepmerge": "4.3.1"
   },
   "devDependencies": {
     "@rollup/plugin-node-resolve": "^15.2.3",
     "@rollup/plugin-typescript": "^11.1.6",
-    "@sentry-internal/eslint-config-sdk": "8.47.0",
-    "@sentry-internal/typescript": "8.47.0",
+    "@sentry-internal/eslint-config-sdk": "8.48.0",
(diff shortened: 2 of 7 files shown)


kpujjigit:
@timfish sounds like this is the PR that will allow ANR events to symbolicate? Would a client on SDK v5.9.0 need to make any changes on their side? Or will it work once this is merged?

Thanks for all you've done!

timfish:
This needs to be merged and then this will be released in v5.10.0

.github/workflows/build.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       # we want that the matrix keeps running, default is to cancel all if one fails.
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
```
chargome, line 58: Why has this changed?

timfish: `ubuntu-latest` changed to Ubuntu 24 and all the Linux tests started failing due to sandbox issues:
https://github.com/getsentry/sentry-electron/actions/runs/12652009781

`The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /home/runner/work/sentry-electron/sentry-electron/.cache/30.5.1-x64/chrome-sandbox is owned by root and has mode 4755.`

We already had to disable the sandbox for older versions of Electron but something has changed in the latest Github Actions images
https://github.com/getsentry/sentry-electron/blob/976b58076faeae4f254234808f1152e6b79f6563/test/e2e/context.ts#L75-L77

We cannot disable the sandbox entirely on Linux because we need to test it's functionality!

I need to find a way to fix this but it's not important for this PR

timfish:
For some reason CI is reporting that:
image

But they have clearly completed sucessfully:

https://github.com/getsentry/sentry-electron/actions/runs/12669658435/job/35307658102?pr=1046

chargome:
@timfish yeah we had this kind of issue before, the branch protection rules are looking for "ubuntu-latest" (exactly that name) but the since the image has changed the name is different. @mydea do you have access rights to modify this?

PR #1053, opened 2025-01-10
fix: Guard against invalid queue entries

### LINKED ISSUE #1051: Error possibly caused by the SDK

- [x] I have checked for existing issues https://github.com/getsentry/sentry-javascript/issues
- [x] I have reviewed the documentation https://docs.sentry.io/
- [x] I am using the latest SDK release https://github.com/getsentry/sentry-javascript/releases

### How do you use Sentry?

Sentry Saas (sentry.io)

### Electron SDK Version

5.8.0

### Electron Version

22.3.27

Windows

### Link to Sentry event

https://desquared.sentry.io/issues/6204868921/?project=4507606472261632

### Steps to Reproduce

The Sentry event I linked does not appear to originate from the customer application's code but rather from our SDK. It points to this file: https://github.com/getsentry/sentry-electron/blob/master/src/main/transports/offline-store.ts

The issue results in a high volume of events being sent.

Zendesk ticket for additional information: https://sentry.zendesk.com/agent/tickets/142036

### Expected Result

The SDK should handle connection timeouts gracefully and not generate excessive error events.

### Actual Result

The SDK generates a large number of error events, causing the customer to reach their error limit.

timfish:
- Closes #1051

I don't know exactly what caused this issue. I suspect it could be due to an invalid read of the data store.

### DIFF
--- a/src/main/transports/offline-store.ts
+++ b/src/main/transports/offline-store.ts
@@ -63,7 +63,7 @@ export function createOfflineStore(userOptions: Partial<OfflineStoreOptions>): O
     });
   }
 
-  function removeStaleRequests(queue: Array<Partial<PersistedRequest> | undefined>): void {
+  function removeStaleRequests(queue: Array<Partial<PersistedRequest>>): void {
     while (queue[0] && isOutdated(queue[0], options.maxAgeDays)) {
       const removed = queue.shift() as PersistedRequest;
       log('Removing stale envelope', removed);
--- a/src/main/transports/offline-store.ts
+++ b/src/main/transports/offline-store.ts
@@ -28,9 +28,9 @@ export interface OfflineStoreOptions {
 
 const MILLISECONDS_PER_DAY = 86_400_000;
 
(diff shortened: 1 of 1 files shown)


src/main/transports/offline-store.ts
```diff
@@ -63,7 +63,7 @@ export function createOfflineStore(userOptions: Partial<OfflineStoreOptions>): O
     });
   }
 
-  function removeStaleRequests(queue: PersistedRequest[]): void {
+  function removeStaleRequests(queue: Array<Partial<PersistedRequest> | undefined>): void {
     while (queue[0] && isOutdated(queue[0], options.maxAgeDays)) {
```
AbhiPrasad, line 67: given we do the `queue[0]` check here, can we enforce that `isOutdated` gets `Partial<PersistedRequest>` instead of `Partial<PersistedRequest> | undefined`?

AbhiPrasad (approved):
Let's ship this as a quick fix for now, but I think we should change the store to see if we can detect scenarios where the data is invalid.

PR #1056, opened 2025-01-13
test: Back to testing on `ubuntu-latest`

timfish:
In a recent PR I had to pin to `ubuntu-22.04` because latest didn't support Electrons sandbox in it's default configuration.

This PR disables the sandbox for the Linux tests and reverts to `ubuntu-latest`. This means we do not test sandbox functionality on Linux but we test other platforms and Electron has extensive tests to cover sandbox behaviour.


timfish:
@mydea Sorry, can you change the merge conditions back to check `unbuntu-latest`?

chargome (approved):

PR #1057, opened 2025-01-14
feat: Explicitly set `user.ip_address: {{auto}}`

timfish:
- Ref https://github.com/getsentry/team-sdks/issues/118

### DIFF
--- a/src/main/integrations/electron-context.ts
+++ b/src/main/integrations/electron-context.ts
@@ -28,7 +28,16 @@ export const electronContextIntegration = defineIntegration(() => {
         delete event.request.headers['User-Agent'];
       }
 
-      const { release = getDefaultReleaseName(), environment = getDefaultEnvironment() } = client.getOptions();
+      const {
+        release = getDefaultReleaseName(),
+        environment = getDefaultEnvironment(),
+        sendDefaultPii,
+      } = client.getOptions();
+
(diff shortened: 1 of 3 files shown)


src/main/integrations/electron-context.ts
```diff
@@ -28,7 +28,16 @@ export const electronContextIntegration = defineIntegration(() => {
...
       }
 
-      const { release = getDefaultReleaseName(), environment = getDefaultEnvironment() } = client.getOptions();
+      const {
+        release = getDefaultReleaseName(),
+        environment = getDefaultEnvironment(),
+        sendDefaultPii,
+      } = client.getOptions();
+
+      if (sendDefaultPii !== false && typeof event.user?.ip_address === 'undefined') {
```
cleptric, line 37: Could we maybe make this more explicit by `sendDefaultPii === true`?

timfish: > A user might opt-out from this behavior by configuring the SDK with send_default_pii: false.

I read this and assumed that `{{auto}}` would be added unless `sendDefaultPii` was explicitly set to false but now I've checked and the default is false!

cleptric: Thanks, reworded the description in the issue to be less confusing 😬

cleptric (approved):

PR #1063, opened 2025-01-20
test: More child process types

timfish:
This PR adds tests for:
- Electron utility process crash
- `child_process.fork` Node process crash
- `child_process.exec` executable crash

Not all platforms can capture child process crashes:

|                    | `event.process`|macOS | Windows | Linux | 
|--------------------|-------|---------|-------|---|
| Electron `utilityProcess`|`utility`|✅|✅|✅ <sup>1</sup>|
| `child_process.fork` | `node`|✅     | ✅       | ❌     |  
| `child_process.exec` | `unknown`|✅     | ❌       | ❌     |  

 - <sup>1</sup> Not working with v29.4.6. Possibly an Electron bug that won't get fixed because v29 is already EOL.

### DIFF
--- a/test/e2e/test-apps/native-sentry/child-exec/event.json
+++ b/test/e2e/test-apps/native-sentry/child-exec/event.json
@@ -1,0 +1,75 @@
+{
+  "method": "envelope",
+  "sentryKey": "37f8a2ee37c0409d8970bc7559c7c7e4",
+  "appId": "277345",
+  "data": {
+    "sdk": {
+      "name": "sentry.javascript.electron",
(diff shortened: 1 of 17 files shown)


AbhiPrasad:
> Not working with v29.4.6

Just with that version? We might want to add it to the docs.

AbhiPrasad (approved):

timfish:
> > Not working with v29.4.6
> 
> Just with that version? We might want to add it to the docs.

Yep, just that single version in failing on Linux! 

Probably worth including the above table in the docs.

PR #1095, opened 2025-03-10
feat: Parse crashpad metadata from minidumps and include in `electron` context

### LINKED ISSUE #1094: Attach information from minidumps as contexts before they are sent to Sentry
In relay we parse minidumps to generate certain contexts: https://github.com/getsentry/relay/blob/0ea0807fe5acd99792b2fb5188de96256d2f4b53/relay-server/src/utils/native.rs#L169C64-L169C78

We should try to do this in the SDK so that folks can mutate it in `beforeSend`, or access the value to set as a tag instead of a context.

timfish:
- Closes #1094 

Previously we pulled the `process_type` value from the minidump via some crude parsing.

This PR adds a full parser to pull annotations from the crashpad stream in the minidumps and includes them in the event `electron` context but prepends `crashpad.` to all the keys. This allows users to use this context in `beforeSend` filters and in other sampling decisions.

Electron includes additional v8 context in the crashpad metadata (like `electron.v8-oom.detail` which gives details about the OOM reason):
https://github.com/electron/electron/blob/d987bee0073310c0bf114dd5e5c9db5e9c68ee59/shell/common/node_bindings.cc#L193

The parser code was mostly aided by the `rust-minidump` crate which has great documentation. I've credited it fully in the source because it would have been a lot more complicated without such a comprehensive library to test against. 

I was going to add unit tests for the parser but that would involve generating minidumps on every platform and adding them to the repository. This has the potential to leak environment variables. Instead, I'm leveraging the e2e tests because this actually tests all supported versions of Electron so we know the parser works across all versions and platforms.

### DIFF
--- a/src/main/integrations/sentry-minidump/minidump-loader.ts
+++ b/src/main/integrations/sentry-minidump/minidump-loader.ts
@@ -77,8 +77,8 @@ export function createMinidumpLoader(getMinidumpPaths: () => Promise<string[]>):
                   data,
                 });
               } catch (e) {
-                const message = e instanceof Error ? e.message : 'Unknown error';
-                logger.warn(`Dropping minidump: ${message}`);
+                const message = e instanceof Error ? e.toString() : 'Unknown error';
+                logger.warn(`Dropping minidump:\n${message}`);
                 break;
               }
 
--- a/src/main/integrations/sentry-minidump/minidump-loader.ts
+++ b/src/main/integrations/sentry-minidump/minidump-loader.ts
@@ -66,20 +66,21 @@ export function createMinidumpLoader(getMinidumpPaths: () => Promise<string[]>):
 
             if (stats.mtimeMs < twoSecondsAgo) {
               const data = await fs.readFile(path);
-              const parsedMinidump = parseMinidump(data);
+              try {
+                const parsedMinidump = parseMinidump(data);
 
-              if (parsedMinidump === undefined) {
-                logger.warn('Dropping minidump as it appears invalid.');
+                logger.log('Sending minidump');
+
+                await callback(parsedMinidump, {
+                  attachmentType: 'event.minidump',
+                  filename: basename(path),
+                  data,
+                });
+              } catch (e) {
+                const message = e instanceof Error ? e.message : 'Unknown error';
+                logger.warn(`Dropping minidump: ${message}`);
                 break;
               }
-
-              logger.log('Sending minidump');
-
-              await callback(parsedMinidump, {
-                attachmentType: 'event.minidump',
-                filename: basename(path),
-                data,
-              });
 
               break;
             }
--- a/src/main/integrations/sentry-minidump/minidump-loader.ts
+++ b/src/main/integrations/sentry-minidump/minidump-loader.ts
@@ -75,14 +75,11 @@ export function createMinidumpLoader(getMinidumpPaths: () => Promise<string[]>):
 
               logger.log('Sending minidump');
 
-              await callback(
-                parsedMinidump,
-                 {
-                  attachmentType: 'event.minidump',
-                  filename: basename(path),
-                  data,
-                },
-              );
+              await callback(parsedMinidump, {
+                attachmentType: 'event.minidump',
+                filename: basename(path),
+                data,
+              });
 
               break;
             }
--- a/src/main/integrations/sentry-minidump/minidump-loader.ts
+++ b/src/main/integrations/sentry-minidump/minidump-loader.ts
@@ -1,9 +1,10 @@
-import { Attachment, basename, logger } from '@sentry/core';
+import { Attachment, logger } from '@sentry/core';
 import { app } from 'electron';
 import { promises as fs } from 'fs';
-import { join } from 'path';
+import { basename, join } from 'path';
 
 import { Mutex } from '../../mutex';
+import { MinidumpParseResult, parseMinidump } from './minidump-parser';
 
 /** Maximum number of days to keep a minidump before deleting it. */
 const MAX_AGE_DAYS = 30;
@@ -13,8 +14,6 @@ const NOT_MODIFIED_MS = 1_000;
 const MAX_RETRY_MS = 5_000;
 const RETRY_DELAY_MS = 500;
 const MAX_RETRIES = MAX_RETRY_MS / RETRY_DELAY_MS;
-
-const MINIDUMP_HEADER = 'MDMP';
 
 function delay(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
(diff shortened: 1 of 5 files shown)


test/e2e/test-apps/native-sentry/renderer/event.json
```diff
@@ -56,7 +56,8 @@
         "version": "{{version}}"
       },
       "electron": {
-        "crashed_url": "app:///src/index.html"
+        "crashed_url": "app:///src/index.html",
+        "crashpad.process_type": "renderer"
```
AbhiPrasad, line 60: I'm surprised the test doesn't add more fields.

src/main/integrations/sentry-minidump/minidump-loader.ts
```diff
@@ -69,19 +65,17 @@ export function createMinidumpLoader(
             const twoSecondsAgo = new Date().getTime() - NOT_MODIFIED_MS;
 
             if (stats.mtimeMs < twoSecondsAgo) {
-              const file = await fs.readFile(path);
-              const data = preProcessFile(file);
+              const data = await fs.readFile(path);
+              const parsedMinidump = parseMinidump(data);
```
AbhiPrasad, line 69: I think we should return an error from `parseMinidump` if we can't parse the minidump and emit debug logging accordingly.

src/main/integrations/sentry-minidump/minidump-parser.ts
```diff
@@ -0,0 +1,256 @@
...
+}
+
+type Location = {
+  dataSize: number;
+  rva: number;
+};
+
+// https://docs.microsoft.com/en-us/windows/win32/api/minidumpapiset/ns-minidumpapiset-minidump_location_descriptor
+function readLocationDescriptor(buf: Buffer, base: number): Location {
+  // pub struct MINIDUMP_LOCATION_DESCRIPTOR {
```
AbhiPrasad, line 70: Let's move this to where the type is being defined above, and add comments about what the field means.

src/main/integrations/sentry-minidump/minidump-parser.ts
```diff
@@ -0,0 +1,256 @@
...
+    version: buf.readUInt32LE(4),
+    streamCount: buf.readUInt32LE(8),
+    streamDirectoryRva: buf.readUInt32LE(12),
+    checksum: buf.readUInt32LE(16),
+    timeDateStamp: new Date(buf.readUInt32LE(20) * 1000),
+    flags: buf.readBigUInt64LE(24),
+  };
+}
+
+type Location = {
```
AbhiPrasad, line 63:
```suggestion
type MinidumpLocation = {
```

src/main/integrations/sentry-minidump/minidump-parser.ts
```diff
@@ -0,0 +1,256 @@
...
+  // pub struct MINIDUMP_MODULE_CRASHPAD_INFO {
+  //   pub version: u32,
+  //   pub list_annotations: MINIDUMP_LOCATION_DESCRIPTOR,
+  //   pub simple_annotations: MINIDUMP_LOCATION_DESCRIPTOR,
+  //   pub annotation_objects: MINIDUMP_LOCATION_DESCRIPTOR,
+  // }
+
+  // const version = buf.readUInt32LE(base)
+  // const list_annotations = readLocationDescriptor(buf, base + 4)
+  // const simple_annotations = readLocationDescriptor(buf, base + 12)
```
AbhiPrasad, line 114: We can remove?

If we're leaving it as docs, let's explicitly comment that.

src/main/integrations/sentry-minidump/minidump-parser.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  try {
+    for (let i = 0; i < header.streamCount; i++) {
+      const stream = readDirectoryStream(buf, header.streamDirectoryRva + i * 12);
+
+      // Crashpad specific stream in Electron minidump files
+      // https://crashpad.chromium.org/doxygen/structcrashpad_1_1MinidumpCrashpadInfo.html
+      if (stream.streamType === 1_129_316_353) {
+        const crashpadInfo = readCrashpadInfoBuffer(buf, stream.location);
+        const crashpadAnnotations = parseCrashpadInfo(buf, crashpadInfo) as CrashpadAnnotations;
```
AbhiPrasad, line 243: can we avoid casting here?

We should validate if `parseCrashpadInfo` works as expected otherwise return `undefined`.

src/main/integrations/sentry-minidump/minidump-parser.ts
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  let header: MinidumpHeader | undefined;
+  try {
+    header = readHeader(buf);
+  } catch (_) {
+    return undefined;
+  }
+
+  if (header.signature !== MINIDUMP_MAGIC_SIGNATURE || buf.length < 10_000) {
```
AbhiPrasad, line 231: Why the check on 10000?

src/main/integrations/sentry-minidump/minidump-parser.ts
timfish: It was here to document the sizes of the fields for the parsing just below it. If it gets moved to the type definition it's just documenting the fields have the same names?

src/main/integrations/sentry-minidump/minidump-parser.ts
timfish: This check existed previously to exclude invalid looking minidumps as reported by users. I'll find the issue and add a link in a comment.

src/main/integrations/sentry-minidump/minidump-parser.ts
AbhiPrasad: I'd rather this be with the type itself, but given this is basically a constructor I see the value in doing it this way. We don't need to change it.

timfish: I'll have another go and copy the rust docs across for the properties because they're useful. I'll document the offsets property rather than just copy the rust structs 🙂

src/main/integrations/sentry-minidump/minidump-loader.ts
```diff
@@ -69,24 +65,23 @@ export function createMinidumpLoader(
...
+                logger.log('Sending minidump');
+
+                await callback(parsedMinidump, {
+                  attachmentType: 'event.minidump',
+                  filename: basename(path),
+                  data,
+                });
+              } catch (e) {
+                const message = e instanceof Error ? e.message : 'Unknown error';
+                logger.warn(`Dropping minidump: ${message}`);
```
AbhiPrasad, line 81: I wonder if we should serialize the entire error to get a stacktrace.

AbhiPrasad (approved):

src/main/integrations/sentry-minidump/minidump-loader.ts
timfish: Do all supported node versions have error.cause?

AbhiPrasad: Yup, we're 18+ now in `@sentry/node`, and [`error.cause` is `16.9.0`](https://nodejs.org/en/blog/release/v16.9.0#error-cause)

PR #1097, opened 2025-03-11
feat: Add `gpuContextIntegration`

### LINKED ISSUE #1091: GPU details as part of crash dump
### Problem Statement

Currently as part of Electron Sentry Crash dump, GPU details are missing. We need GPU information inorder to find out, if the user is not on latest driver version.

From the following documentation:
https://develop.sentry.dev/sdk/data-model/event-payloads/contexts/#gpu-context
GPU details are part of payloads, but:
https://github.com/getsentry/sentry-javascript/blob/master/packages/node/src/integrations/context.ts
Above code doesn't have the GPU details.

How to enabled/collect the GPU details as per above suggestion?
Also with respect to multiple GPUs, will GPU details of all the GPUs present in system be captured?

Thanks,
Venkatesan

### Solution Brainstorm

Make use of:
https://develop.sentry.dev/sdk/data-model/event-payloads/contexts/#gpu-context
and
https://github.com/getsentry/sentry-javascript/blob/master/packages/node/src/integrations/context.ts

or should the app collect these details as part of custom data.

timfish:
- Closes #1091

If you want full GPU information to be included, you should add the integration with `infoLevel: 'complete'`:
```ts

Sentry.init({
  dsn: '__DSN__',
  integrations: [gpuContextIntegration({ infoLevel: 'complete' })],
});
```

image

### DIFF
--- a/src/main/integrations/gpu-context.ts
+++ b/src/main/integrations/gpu-context.ts
@@ -1,0 +1,76 @@
+import { defineIntegration, dropUndefinedKeys, Event } from '@sentry/core';
+import { app } from 'electron';
+
+interface Options {
+  infoLevel: 'basic' | 'complete';
+}
+
+interface GpuContext extends Record<string, unknown> {
+  name: string;
+  active?: boolean;
+  vendor_id?: string;
+  vendor_name?: string;
+  device_id?: string;
+  driver_version?: string;
+}
+
+interface GpuDevice {
+  active: boolean;
+  deviceId: number;
+  deviceString?: string;
+  driverVendor?: string;
+  driverVersion?: string;
+  vendorId: number;
+  vendorString?: string;
+}
+
+interface GpuInfoResult {
+  gpuDevice: GpuDevice[];
+}
+
+function gpuDeviceToGpuContext(device: GpuDevice): GpuContext {
+  return dropUndefinedKeys({
+    name: device.deviceString || 'GPU',
+    active: device.active,
+    vendor_id: `0x${device.vendorId.toString(16)}`,
+    vendor_name: device.vendorString,
+    device_id: `0x${device.deviceId.toString(16)}`,
+    driver_version: device.driverVersion,
+  });
+}
+
+/**
+ * Adds additional Electron context to events
+ */
+export const gpuContextIntegration = defineIntegration((options: Options = { infoLevel: 'basic' }) => {
+  let gpuContextsPromise: Promise<GpuContext[]> | undefined;
+
+  return {
+    name: 'GpuContext',
+    setup() {
+      app.on('gpu-info-update', async () => {
+        gpuContextsPromise = (app.getGPUInfo(options.infoLevel) as Promise<GpuInfoResult>).then((result) =>
+          result.gpuDevice.map(gpuDeviceToGpuContext),
+        );
+      });
+    },
+    processEvent: async (event): Promise<Event> => {
+      if (gpuContextsPromise) {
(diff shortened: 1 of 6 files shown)


src/main/integrations/gpu-context.ts
```diff
@@ -0,0 +1,76 @@
...
+    setup() {
+      app.on('gpu-info-update', async () => {
+        gpuContextsPromise = (app.getGPUInfo(options.infoLevel) as Promise<GpuInfoResult>).then((result) =>
+          result.gpuDevice.map(gpuDeviceToGpuContext),
+        );
+      });
+    },
+    processEvent: async (event): Promise<Event> => {
+      if (gpuContextsPromise) {
+        const gpuContexts = await gpuContextsPromise;
```
AbhiPrasad, line 59: shouldn't we cache the resolved promise to avoid having to tick the event loop again?

src/main/integrations/gpu-context.ts
```diff
@@ -0,0 +1,76 @@
...
+
+interface GpuInfoResult {
+  gpuDevice: GpuDevice[];
+}
+
+function gpuDeviceToGpuContext(device: GpuDevice): GpuContext {
+  return dropUndefinedKeys({
+    name: device.deviceString || 'GPU',
+    active: device.active,
+    vendor_id: `0x${device.vendorId.toString(16).padStart(4, '0')}`,
```
AbhiPrasad, line 35: Why do this transform?

src/main/integrations/gpu-context.ts
```diff
@@ -0,0 +1,76 @@
...
+  device_id?: string;
+  driver_version?: string;
+}
+
+interface GpuDevice {
+  active: boolean;
+  deviceId: number;
+  deviceString?: string;
+  driverVendor?: string;
+  driverVersion?: string;
```
AbhiPrasad, line 22: I think `driverVersion` maps to `version` on the GPU context.

src/main/integrations/gpu-context.ts
```diff
@@ -0,0 +1,76 @@
+import { defineIntegration, dropUndefinedKeys, Event } from '@sentry/core';
+import { app } from 'electron';
+
+interface Options {
+  infoLevel: 'basic' | 'complete';
```
AbhiPrasad, line 5: Could we add docs for this? We can link to https://www.electronjs.org/docs/latest/api/app#appgetgpuinfoinfotype

AbhiPrasad (commented):
Could you share an example of in Sentry of what a crash with this GPU context looks like?

src/main/integrations/gpu-context.ts
timfish: Device and vendor IDs are often referred to in hex notation.

For example the default windows display driver vendor ID is 0x1414. 

https://devicehunt.com/view/type/pci/vendor/1414

Saves users from having to the conversion before searching!

src/main/integrations/gpu-context.ts
timfish: I thought that might be the case until I saw the example which has `metal`:
https://develop.sentry.dev/sdk/data-model/event-payloads/contexts/#example-gpu-context

timfish: @bruno-garcia do have any insight for this `version` property in the GPU context?

timfish:
> Could you share an example of in Sentry of what a crash with this GPU context looks like?

Added an image to the top post with the GPU context as displayed in Sentry

PR #1117, opened 2025-03-31
fix: Missing DSC envelope header for renderer tracing

### LINKED ISSUE #1114: Missing `sample_rate` in dynamic sampling context
Found events on `sentry.javascript.electron` version `5.7.0` that are missing `sample_rate` in the dynamic sampling context. I'm not sure if this has been fixed in more recent SDK versions!

EAP relies on client_sample_rate for span metrics extrapolation. Relay currently [extracts](https://github.com/getsentry/relay/blob/master/relay-server/src/services/processor/span/processing.rs#L306-L331) this data from the sample_rate field in the dynamic sampling context. But we noticed that in the Electron SDK, this data was simply not available. Which means it isn't extracted properly in relay and extrapolation isn't working as expected for these events.

This notion doc has more context and links to events (contains some sensitive information, so don't want to share here)
https://www.notion.so/sentry/Missing-sample_rate-in-dynamic-sampling-context-for-electron-sdk-1c28b10e4b5d8063b47ccfe6e1c60686

timfish:
- Closes #1114

Events and transactions from renderer (browser) processes are re-captured through the main (Node.js) process so that they include the same full app context as all other events. However, this meant that the `trace` context in the envelope header from the renderer was dropped and it always contained the DSC for the main process. If there was no active span in the main process, `sample_rate` would be missing from the DSC. Even if there was an active span in the main process, the `sample_rate` and other context would be for the main process.

This PR pulls the `trace` context from the envelope header from the renderer and passes it back through the event pipeline via `sdkProcessingMetadata` which is where it is pulled from when creating the envelope headers:
https://github.com/getsentry/sentry-javascript/blob/7f1087d3fb6a2be0aa7427c49d37dd352dba07da/packages/core/src/utils-hoist/envelope.ts#L254

All other envelope types (other than event/transaction) are sent complete without changes so only these types of envelopes require the pass through of DSC.

### DIFF
--- a/test/e2e/test-apps/other/browser-tracing/event.json
+++ b/test/e2e/test-apps/other/browser-tracing/event.json
@@ -112,5 +112,8 @@
       "event.environment": "javascript",
       "event.origin": "electron"
     }
+  },
+  "dynamicSamplingContext": {
+    "sample_rate": "1"
   }
 }
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -1,4 +1,4 @@
-import { Attachment, Event, logger, parseEnvelope, ScopeData, SentryError } from '@sentry/core';
+import { Attachment, DynamicSamplingContext, Event, logger, parseEnvelope, ScopeData, SentryError } from '@sentry/core';
 import { captureEvent, getClient, getCurrentScope } from '@sentry/node';
 import { app, ipcMain, protocol, WebContents, webContents } from 'electron';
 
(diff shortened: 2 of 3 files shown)


test/e2e/test-apps/other/browser-tracing/event.json
```diff
@@ -112,5 +112,8 @@
       "event.environment": "javascript",
       "event.origin": "electron"
     }
+  },
+  "dynamicSamplingContext": {
+    "sample_rate": "1"
```
timfish, line 117: Here we modify the browser tracing e2e test to ensure the `sample_rate` is indeed in the DSC

AbhiPrasad:
yeah let's backport this

PR #1125, opened 2025-04-04
feat: Optionally capture manufacturer and model device context

### LINKED ISSUE #1112: Manufacturer and model details as part of crash dump
### Problem Statement

On Windows crashes, system manufacturer and model details are missing in crash dump.

Image

These details needs to be present as part of device details.

### Solution Brainstorm

There are windows system api's to collect these info, when CPU description is fetched, along with this System Manufacturer and Model can be fetched.

timfish:
- Closes #1112 

This PR adds to the `additionalContextIntegration` so you can optionally include the manufacturer and model in the device context:

```ts

Sentry.init({
  dsn: '__DSN__',
  integrations: [Sentry.additionalContextIntegration({ deviceModelManufacturer: true })]
});
```

It's worth noting that testing this on my Windows laptop, pulling this data via powershell takes around 400ms. This is a one-off cost as the result is cached. This data is captured asynchronously but it will delay sending of the first event.  

image

### DIFF
--- a/src/main/integrations/additional-context.ts
+++ b/src/main/integrations/additional-context.ts
@@ -11,7 +11,8 @@ export interface AdditionalContextOptions {
    */
   screen: boolean;
   /**
-   * Capture device model and manufacturer on Windows
+   * Capture device model and manufacturer.
+   * Only supported in Windows.
    * @default false
    */
   deviceModelManufacturer: boolean;
--- a/src/main/integrations/additional-context.ts
+++ b/src/main/integrations/additional-context.ts
@@ -12,7 +12,7 @@ export interface AdditionalContextOptions {
   screen: boolean;
   /**
    * Capture device model and manufacturer.
-   * Only supported in Windows.
+   * No supported on Linux.
    * @default false
    */
   deviceModelManufacturer: boolean;
@@ -48,8 +48,9 @@ function getWindowsDeviceModelManufacturer(): Promise<DeviceContext> {
               return;
             }
           } catch (_) {
-            resolve({});
+            //
           }
+          resolve({});
         },
       );
     } catch (_) {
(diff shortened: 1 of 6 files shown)


src/main/integrations/additional-context.ts
```diff
@@ -1,28 +1,75 @@
...
 import { mergeEvents } from '../merge';
 
 export interface AdditionalContextOptions {
+  /**
+   * Capture dimensions and resolution of the primary display
+   * @default true
+   */
   screen: boolean;
+  /**
+   * Capture device model and manufacturer on Windows
```
AbhiPrasad, line 14:
```suggestion
   * Capture device model and manufacturer.
   * Only supported in Windows.
```

AbhiPrasad (approved):
How hard is it to get this info for other platforms? Or does it already show up there?

AbhiPrasad:
I wonder if we can ask the https://github.com/getsentry/sentry-cocoa people for help here, they must run into this for macos stuff

timfish:
It looks like Sentry already does conversions between iOS device IDs and names and it looks like this might soon include Mac models too. 
- https://github.com/getsentry/sentry/issues/68258

I'm going to add macOS support in this PR and initially these will just be displayed as the model Id (`MacBookPro18,3`) and then later we'll likely get the Apple "marketing names".

kahest:
@timfish thanks for adding this! can you test if this also works with sentry-cocoa events, e.g. [some of these](https://sentry-sdks.sentry.io/issues/?project=5428557&query=device%3A%2Amacbook%2A%20sdk.name%3Asentry.cocoa%2A&referrer=issue-list&sort=date&statsPeriod=90d&viewId=103277)?

timfish:
I suspect one the device mapping get added/updated in Sentry, both for Electron and Cocoa events, the device model will get a more useful name. (ie. `MacBookPro18,3` -> `MacBook Pro (M1 Pro, 14-inch, 2021)`)

timfish:
Unfortunately we can only test this on macOS because this kind of OEM data doesn't appear to be set in GHA CI agents

AbhiPrasad (approved):

PR #1132, opened 2025-04-07
test: Pin webpack version because v5.99.0 appears broken

timfish:


AbhiPrasad (approved):

AbhiPrasad (changes requested):
Actually it seems like https://github.com/webpack/webpack/releases/tag/v5.99.2 has fixed it, let's try again with that?

PR #1159, opened 2025-05-26
feat: Add structured logging support

### LINKED ISSUE #1154: Sentry Structured Logging for Electron
https://develop.sentry.dev/sdk/telemetry/logs/

Sentry is adding support for [structured logging](https://github.com/getsentry/sentry/discussions/86804). Let's add it to the Electron SDK!

1. Add the Public API (SDK options and methods) as per the [docs](https://develop.sentry.dev/sdk/telemetry/logs/#public-api)

You can look at the Node SDK for how to structure this: https://github.com/getsentry/sentry-javascript/blob/develop/packages/node/src/logs/exports.ts

```ts
const { info, warn, fmt } = Sentry.logger;

info(fmt`Updated profile for user ${userId}`);
warn(fmt`Rate limit approaching for endpoint ${endpoint}. Requests: ${requests}, Limit: ${limit}`);

// printf based templating using Node.js util.format
info('User %s logged in successfully', [123]);
warn('Failed to load user %s data', [123], { errorCode });
```

We might not be able to support printf based templating in the renderer process unless we IPC down the template to use `util.format` in the main thread.

2. Make sure the the SDK follow the [documented behavior](https://develop.sentry.dev/sdk/telemetry/logs/#sdk-behavior)

You can look at `server-runtime-client` in the monorepo to structure a buffering strategy: https://github.com/getsentry/sentry-javascript/blob/79480cff1f08079a8c09e10e2049c9a0f2c151ec/packages/core/src/server-runtime-client.ts#L52.

3. Think about what logging integrations we should add that are electron specific. For now we don't duplicate breadcrumbs, we will evaluate that at a later time.

4. After we release this, we should create a GH discussion that contains instructions for setting up the SDK in your repo. Eventually this will be moved into the primary docs. See the [JS SDK's GH discussion](https://github.com/getsentry/sentry-javascript/discussions/15916) for inspiration.

timfish:
- Closes #1154

This PR requires:
- A JavaScript SDK release with these:
  - https://github.com/getsentry/sentry-javascript/pull/16352
  - https://github.com/getsentry/sentry-javascript/pull/16387
- ~~Tests!~~

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -19,6 +19,7 @@ import { rendererProfileFromIpc } from './integrations/renderer-profiling';
 import { mergeEvents } from './merge';
 import { normalizeReplayEnvelope } from './normalize';
 import { ElectronMainOptionsInternal } from './sdk';
+import { SDK_VERSION } from './version';
 
 let KNOWN_RENDERERS: Set<number> | undefined;
 let WINDOW_ID_TO_WEB_CONTENTS: Map<string, number> | undefined;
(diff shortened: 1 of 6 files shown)


AbhiPrasad (commented):
How does `beforeSendLog` work here?

timfish:
> So adding beforeSendLog to the main process only affects the logs captured via the main process?

Yes.

For renderer events, `beforeSend` is called in both the renderer and main process because they are captured in both.

The issue with `beforeSendLog` is that it's passed a `Log` type whereas the main process only deals with `SerializedLog` from renderers. If we are to call `beforeSendLog` for renderer logs we'd need to convert them back to `Log`.

AbhiPrasad (approved):

Issue #1150, opened 2025-05-01
A way to stop ANR manually


Galkon:
### Problem Statement

Every now and then I see ANR issues related to objects being destroyed during app quit. I'm not sure what is expected behavior by Sentry SDK upon app quit, but given the varying ways people can prevent default behavior for `before-quit`, `will-quit`, `window-all-closed`, etc -- perhaps an explicit way to stop ANR before shut down would be ideal.

### Solution Brainstorm

I tried this, based on the source code:
```Sentry.getClient()?.getIntegrationByName('Anr')?.stopWorker()```

But `getClient()` is undefined, for some reason? So I'm not sure if I can accomplish this at all currently. As well as `stopWorker()` is not exposed via types.

I would suggest something like `Sentry.disableAnrIntegration()` or something easy for users to call from main or renderer processes.

**Comment by AbhiPrasad** [comment]
Hey @Galkon thanks for writing in! Do you mind sharing a link to the ANR issue? I'm curious what the stacktrace looks like.

Maybe we need to just shut the worker off during exit, we already have similar logic when we suspend or lock the screen: https://github.com/getsentry/sentry-electron/blob/abcee88fcf2d61d222840be9cd6d15ec6012d64c/src/main/integrations/anr.ts#L28-L42

**Comment by Galkon** [comment]
Thanks for the quick response @AbhiPrasad. Here is the issue I am seeing in main process logs upon app exit -- not _always_ but sometimes:
```
Unhandled rejection: TypeError: Object has been destroyed
    at /Applications/Highlight.app/Contents/Resources/app.asar/node_modules/@sentry/electron/main/anr.js:63:25
    at /Applications/Highlight.app/Contents/Resources/app.asar/node_modules/@sentry/electron/main/anr.js:101:21
    at Timeout._onTimeout (/Applications/Highlight.app/Contents/Resources/app.asar/node_modules/@sentry/core/build/cjs/utils-hoist/anr.js:30:9)
    at listOnTimeout (node:internal/timers:581:17)
    at process.processTimers (node:internal/timers:519:7)
```

In my case, I also granularly control `before-quit` event behavior, so I would be curious to see how you might solve stopping the worker during exit in main process. I image in a renderer process, this would be in something like `window.addEventListener('beforeunload', () => { ... })`

AbhiPrasad:
ah is this from an older sdk version? What electron and SDK version are you using?

Galkon:
[...]

PR #1165, opened 2025-06-03
fix: Don't capture stack traces from destroyed renderers

timfish:
- Reported here https://github.com/getsentry/sentry-electron/issues/1150#issuecomment-2936945459

We should check if the webContents has been destroyed before attempting to contact the debugger.


AbhiPrasad (approved):
Needs a test - but not blocking merge on this

(we can also probably sneak this into the release)

timfish:
This is going to be tricky to test for as I'm not 100% sure how to reproduce it. I'll try a few things!

AbhiPrasad:
I would just merge this in and release. I'd like to get the structured logs work out!

PR #1176, opened 2025-06-26
chore: Update development dependencies

timfish:
Also deleted the lock file and started from scratch!

Are there any issues with using the latest TypeScript?


timfish:
It looks like the rollup build fails to exit intermittently on macOS. I've seen this locally occasionally before this PRs changes but have no idea what causes it!

PR #1188, opened 2025-07-16
feat: Add `eventLoopBlockIntegration`

### LINKED ISSUE #1180: Test `eventLoopBlockIntegration` with Electron

timfish:
- Closes #1180

This PR adds an `eventLoopBlockIntegration` which wraps the node integration while adding the additional tags and enable/disable features we require in Electron. 

You need to add `@sentry/node-native` to your dependencies and then import the integration from `@sentry/electron/native`. Using a seperate export means the optional peer dependency and native module won't be picked up by users who aren't using this feature.

 ```js
 import { eventLoopBlockIntegration } from '@sentry/electron/native';
 
 Sentry.init({
   dsn: '__YOUR_DSN__',
   integrations: [
     eventLoopBlockIntegration({
       threshold: 500, // Report blocks longer than 500ms
     }),
   ],
 });
 ```


AbhiPrasad (approved):
This only works in the main thread right? Let's make that clear in the doc string (or maybe we can throw an error if it's being used in the renderer process?)

timfish:
> This only works in the main thread right? 

It might work with renderer processes when Node integration is enabled but that feature has been discouraged for years and nobody should really be using it nowadays. 

I'll ensure we throw if it's loaded in the renderer!

AbhiPrasad (approved):

AbhiPrasad (approved):

PR #1202, opened 2025-07-31
test: Fix e2e flakes

timfish:
Testing in CI...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,7 +73,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -73,7 +73,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-22.04]
```
AbhiPrasad, line 76: doesn't this mean we don't test on windows or macos anymore?

AbhiPrasad (approved):

PR #1203, opened 2025-07-31
fix: Set TypeScript `moduleResolution` to `node16`

### LINKED ISSUE #1200: Error when updating `moduleResolution` with tsc
@tmm1 said:

> hit this when i tried to update moduleResolution
>
> ```$ tsgo -p src --noEmit --incremental false
> node_modules/@sentry/electron/esm/main/index.d.ts:4:48 - error TS2835: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Did you mean './integrations/electron-breadcrumbs.js'?
> 
> 4 export { electronBreadcrumbsIntegration } from './integrations/electron-breadcrumbs';
>                                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```

timfish:
- Closes #1200

Although maybe we should change this here?
https://github.com/getsentry/sentry-javascript/blob/5bf18c47f775ca29683f03122bacab7616d92020/packages/typescript/tsconfig.json#L11

### DIFF
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -11,6 +11,7 @@
     "rootDir": "src",
     "target": "ES2020",
     "esModuleInterop": true,
-    "ignoreDeprecations": "5.0"
+    "ignoreDeprecations": "5.0",
+    "moduleResolution": "node16"
   }
 }


tsconfig.build.json
```diff
@@ -11,6 +11,7 @@
     "rootDir": "src",
     "target": "ES2020",
     "esModuleInterop": true,
-    "ignoreDeprecations": "5.0"
+    "ignoreDeprecations": "5.0",
+    "moduleResolution": "node16"
```
AbhiPrasad, line 15: would this be considered a breaking change?

src/main/context.ts
```diff
@@ -2,7 +2,7 @@
 import { Event, EventHint, SdkInfo } from '@sentry/core';
 import { NodeClient } from '@sentry/node';
 import { app } from 'electron';
-import { SDK_VERSION } from './version';
+import { SDK_VERSION } from './version.js';
```
AbhiPrasad, line 5: does this get enforced via linting? If not, we should lint for this somehow.

also, do we also have to do this for the node and browser sdks? because that's what electron uses under the hood.

timfish: I couldn't reproduce this issue with the Node SDK and I couldn't work out why it doesn't affect it. 

When we migrate the package to be ESM first (ie. `"type": "module"` in the package.json I think this will be a requirement for all packages.

I'll look into linting!

tsconfig.build.json
timfish: I don't think this impacts the output code at all so don't think it's a breaking change.

PR #1215, opened 2025-08-15
chore: remove @rollup/plugin-node-resolve, use npm-run-all2

outslept:
`npm-run-all` is no longer maintained and can be vulnerable, `npm-run-all2` is a maintained drop-in replacement (fork).

regarding `rollup/plugin-node-resolve`. I was running this to investigate:

```sh
git log --all -p -S "@rollup/plugin-node-resolve"
```

The plugin was added in #756 but was never actually used in `rollup.config.mjs`.
It was added to `package.json` but overlooked in the actual configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "koa-bodyparser": "^4.3.0",
     "koa-tree-router": "^0.12.1",
     "latest-version": "^7.0.0",
-    "npm-run-all": "^4.1.5",
+    "npm-run-all2": "^8.0.4",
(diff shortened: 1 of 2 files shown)


outslept:
rollup/plugin-node-resolve should have no effect until declared and npm-run-all definitely shouldn't make the marks go red. Fails should be unrelated to the changes, but I may be wrong

timfish:
We've got some flakey tests on Ubuntu most likely caused by it running out of resources.

PR #1239, opened 2025-09-18
test: Shard E2E Tests

timfish:
This PR:
- Shards the e2e tests into 4 shards per Electron version
- Copies the packed tgz between CI jobs so less re-building occurs in each job
- Disables a couple of tests on Linux
- Moves a unit test requiring full build output to post build checks
- Fixes a test that was intermittent


timfish:
@AbhiPrasad I think we need to merge this one first if we are to have any chance of getting anything through CI 🫠

.github/workflows/build.yml
```diff
@@ -73,8 +71,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-2022, macos-latest]
```
AbhiPrasad, line 74: is there a date when `windows-2022` will no longer be available?

AbhiPrasad (approved):

PR #1260, opened 2025-11-17
feat: Add support for Sentry Metrics

AbhiPrasad:
resolves https://github.com/getsentry/sentry-electron/issues/1263

https://docs.sentry.io/product/explore/metrics/

https://linear.app/getsentry/project/metrics-electron-b67f839c2d98/overview


timfish (approved):
LGTM if you can get the tests passing. 

Looks like the SDK is not being overridden to Electron?

AbhiPrasad:
I wonder if these timeouts are related or just flakes 🤔

timfish:
My guess would be timeouts since it's not impacting every os/version combo

timfish (approved):

Issue #1264, opened 2025-11-24
Add ui profiling support to electron


AbhiPrasad:
https://develop.sentry.dev/sdk/telemetry/profiles/continuous-profiling-api/

PR #1267, opened 2025-11-25
feat: Add support for UI profiling

AbhiPrasad:
resolves https://github.com/getsentry/sentry-electron/issues/1264

This PR adds support for UI profiling to Electron.

The `profilesSampleRate` is now deprecated in favor of `profileSessionSampleRate`. The decisions if something should be profiled or not will now be made at SDK initialization (start of a user session on page load). And then, the whole user session will be continuously profiled.

In `profileLifecycle: "manual"` (default), users can start and stop profiling with `Sentry.uiProfiler.startProfiler()` and `.stopProfiler()`.

In `profileLifecycle: "trace"` the SDK automatically profiles when there is at least one active root/segment span. This is most similar to our previous transaction-based approach.


AbhiPrasad:
hmm @timfish any idea why the renderer e2e tests are failing here?

timfish:
Looks like flake since they passed on the third attempt 🤔

AbhiPrasad:
alright cool! This is ready for review now then. @timfish if you have a sufficiently complicated electron app, would be good to test this on that. My apps are all pretty simple so I don't know if there are cases where this could break.

PR #1280, opened 2026-01-02
ci: Simpler Electron version lookup

timfish:
Previously we used `electron-latest-versions` to get the latest Electron versions but the size of the Electron release metadata has grown substantially since this was added years ago and it now fails to run in CI because the download times out.

This PR changes to use `npm view electron dist-tags --json` to list the tags which can then be iterated through to get the latest version of each major.

### DIFF
--- a/scripts/update-electron-versions.mjs
+++ b/scripts/update-electron-versions.mjs
@@ -1,5 +1,10 @@
 import { spawnSync } from 'node:child_process';
+import { writeFileSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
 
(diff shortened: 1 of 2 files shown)


scripts/update-electron-versions.mjs
```diff
@@ -0,0 +1,23 @@
...
+
+  if (version) {
+    versions.push(version);
+    startVersion++;
+  } else {
+    break;
+  }
+}
+
+console.log(JSON.stringify(versions));
```
andreiborza: l: Maybe we can just write the json file directly here too instead of printing and piping the output in the package.json?

Feel free to disregard tho.

andreiborza (approved):

andreiborza (approved):
LGTM

PR #1282, opened 2026-01-07
feat: Add `startupTracingIntegration`

timfish:
Used this like:

`main.js`
```ts

Sentry.init({
  dsn: '__DSN__',
  tracesSampleRate: 1,
  integrations: [Sentry.startupTracingIntegration()],
});
```
The `startupTracingIntegration` captures Electron startup events as tracing spans.

If you also use the `browserTracingIntegration` in the renderer, this envelope is intercepted and it's spans are added to the main process startup transaction.

`renderer.js`
```ts
import * as Sentry from '@sentry/electron/renderer';

Sentry.init({
  integrations: [Sentry.browserTracingIntegration()],
  tracesSampleRate: 1,
});
```

## Outstanding issues
- https://github.com/getsentry/sentry/pull/105903

Once this is merged it will look like this:

image

### DIFF
--- a/src/main/integrations/startup-tracing.ts
+++ b/src/main/integrations/startup-tracing.ts
@@ -32,8 +32,8 @@ function rootTransaction(): Span {
 
     startSpanManual(
       {
-        name: 'electron.startup',
-        op: 'auto.electron.startup',
+        name: 'Startup',
+        op: 'app.start',
         startTime,
         attributes: {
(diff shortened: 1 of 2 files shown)


timfish:
> It would be nice to enable this by default, but we can probably do that in the next major.

Until then I could add it to the docs config selector like we already do for browser tracing.

PR #1303, opened 2026-01-27
fix: Replace deprecated url.parse with parseStringToURLObject

AbhiPrasad:
Node.js 24+ (Electron 40+) emits deprecation warnings for url.parse() which get captured by captureConsoleIntegration and cause test failures.

Replace url.parse() calls in net-breadcrumbs.ts with parseStringToURLObject from @sentry/core which uses the WHATWG URL API.

Also updates e2e test versions to include Electron 40.0.0.

### DIFF
--- a/test/e2e/test-apps/native-sentry/renderer-limit/src/main.js
+++ b/test/e2e/test-apps/native-sentry/renderer-limit/src/main.js
@@ -23,7 +23,9 @@ app.on('ready', () => {
 
   // Keep reloading the window to cause multiple crash events
   app.on('render-process-gone', () => {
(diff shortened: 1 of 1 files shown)


timfish:
Could the remaining `url.parse` `DeprecationWarning` be due to Yarn v1? I've seen this during development. Maybe we just loosen the e2e test because this shouldn't cause the test to fail.

timfish:
It comes internally from Electron:
https://github.com/electron/electron/issues/49550

timfish (approved):

timfish:
I modified this PR to get the test passing with the latest version of Electron

Issue #1043, opened 2024-12-27
Integrating sentry electron using ESM: regression caused for electron-updater: require is not defined


CatStudioApp:
### Is there an existing issue for this?

- [x] I have checked for existing issues https://github.com/getsentry/sentry-javascript/issues
- [x] I have reviewed the documentation https://docs.sentry.io/
- [x] I am using the latest SDK release https://github.com/getsentry/sentry-javascript/releases

### How do you use Sentry?

Self-hosted/on-premise

### Electron SDK Version

5.9.0

### Electron Version

33.2.1

### What platform are you using?

MacOS

### Steps to Reproduce

We're using this template: https://github.com/electron-vite/electron-vite-react.git

```ts
import * as Sentry from "@sentry/electron/main";

export async function initSentry() {
  Sentry.init({
    debug: import.meta.env.DEV,
    environment: import.meta.env.DEV ? "development" : "production",
    dsn: "xxx",
  });
}
```

btw, this can work around this bug:

```ts
import { createRequire } from "node:module";

const require = createRequire(import.meta.url);

export function initSentry() {
  const Sentry = require("@sentry/electron/main");
  const log = require("electron-log");
  Sentry.init({
    debug: true,
    environment: import.meta.env.DEV ? "development" : "production",
    dsn: "xxx",
  });
  log.info("Sentry initialized");
}

initSentry();

```

### Expected Result

everything should work

### Actual Result

electron-updater stopped from working after integrating sentry electron.
`require is not defined in ES module scope, you can use import instead`
`This file is being treated as an ES module because it has a '.js' file extension`

timfish:
Sorry, you will need to supply a more complete example because I can't reproduce this issue when I add Sentry to the above repository.

For reference, I checked out the above repo and made these changes and don't get any build or runtime errors:
[electron-vite-react-c56d5f3-Sentry-working.patch](https://github.com/user-attachments/files/18285373/electron-vite-react-c56d5f3-Sentry-working.patch)

timfish:
We didn't hear back from you but feel free to re-open this issue if it can still be reproduced!

Issue #1007, opened 2024-10-31
Add to docs how to init sentry-electron when using a bundled electron ESM main thread


tom2strobl:
### Is there an existing issue for this?

- [x] I have checked for existing issues https://github.com/getsentry/sentry-javascript/issues
- [x] I have reviewed the documentation https://docs.sentry.io/
- [x] I am using the latest SDK release https://github.com/getsentry/sentry-javascript/releases

### How do you use Sentry?

Sentry Saas (sentry.io)

### Electron SDK Version

5.6.0

### Electron Version

32.1.2

### What platform are you using?

MacOS

### Steps to Reproduce

Preface: to my knowledge one should [not use top level await](https://x.com/MarshallOfSound/status/1737128216861483194?s=20) to avoid electron deadlocks with the ready event

I'm aware that sentry needs load and initialize before electron, and I saw [the recommendation to await generously](https://www.electronjs.org/docs/latest/tutorial/esm#caveats) so my first intention was:

`main.ts` that gets electron cli'ed (minimal reproduction)

```javascript
async function initializeSentry() {
  const Sentry = await import('@sentry/electron/main')
  Sentry.init({ dsn: process.env.SENTRY_DSN })
}

async function bootApp() {
  await initializeSentry()
  const { app } = await import('electron')
  await app.whenReady()
  console.log('we are ready')
}

bootApp()
```

results in `SentryError: Sentry SDK should be initialized before the Electron app 'ready' event is fired`

So I tried to move up everything:

```javascript
import * as Sentry from '@sentry/electron/main'
Sentry.init({ dsn: process.env.SENTRY_DSN })

async function bootApp() {
  await initializeSentry()
  const { app } = await import('electron')
  await app.whenReady()
  console.log('we are ready')
}

bootApp()
```

which now results in some cryptic hacky opentelementry require business

```
TypeError: Cannot read properties of undefined (reading 'require')
    at new Hook (file:///<somepath>/main.bundle.dev.js:37874:40)
    at __webpack_modules__.../node_modules/@opentelemetry/instrumentation/build/esm/platform/node/RequireInTheMiddleSingleton.js.RequireInTheMiddleSingleton._initialize (file:///<somepath>/main.bundle.dev.js:21426:9)
    at new RequireInTheMiddleSingleton (file:///<somepath>/main.bundle.dev.js:21422:14)
[...]

PR #1354, opened 2026-04-15
test: add breaking electron-vite esm example

biw:
Adds a new `examples/electron-vite-esm` fixture to reproduce the current ESM failure path in an `electron-vite` app.

This PR:
- adds a minimal Electron + Vite + Vue example with `"type": "module"`
- initializes Sentry in both the Electron main process and renderer
- loads `electron-squirrel-startup` via `createRequire(import.meta.url)` in the ESM main entry
- adds an e2e test that runs with `skipEsmAutoTransform: true` and asserts the captured fatal event

This example will fail until https://github.com/open-telemetry/opentelemetry-js/pull/6590 is merged in, but I think there's value in adding it here since #1043 and #1007 seem like the same underlying issue. Maybe after https://github.com/open-telemetry/opentelemetry-js/pull/6590 is merged in?

### DIFF
--- a/examples/electron-vite-esm/electron.vite.config.mjs
+++ b/examples/electron-vite-esm/electron.vite.config.mjs
@@ -1,0 +1,17 @@
+import { resolve } from 'path';
+import { defineConfig } from 'electron-vite';
+import vue from '@vitejs/plugin-vue';
+
+export default defineConfig({
+  main: {
+    plugins: [],
+  },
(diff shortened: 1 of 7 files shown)


biw:
@timfish maybe I shouldn't have referenced #1007 in my PR 😅

timfish:
Feel free to open an issue if you have a minimal preproduction!

PR #1372, opened 2026-05-19
chore: Modernize TypeScript configuration

timfish:
This PR:
- Moves to TypeScript v6
- Fixes a couple of type issues highlighted by the above change
- Moves to a single, simplified, self-contained, modern `tsconfig.json`


logaretm (approved):
Nice

PR #1389, opened 2026-06-18
fix(net): Propagate correct distributed tracing IDs

lcian:
The previous code `getTraceData({ span, propagateTraceparent })` could cause `0...0` Trace and Span IDs to be propagated when there's no parent span (either TwP i.e. `tracesSampleRate == null`, or `tracesSampleRate > 0` but simply no span around the HTTP call).
`span` is created with `onlyIfParent: true`, so when there's no parent, `span == NonRecordingSpan` with `INVALID_SPAN_CONTEXT`, and calling `getTraceData` returns `0...0` trace and span IDs that are propagated.

This PR mirrors the code in https://github.com/getsentry/sentry-javascript/blob/ccf5571c78c1116ba15fa7186234c86dcf57db5a/packages/core/src/integrations/http/client-subscriptions.ts#L119 to fix said behavior.
It stops passing `span` explicitly, so that `getTraceData` can either detect the active span injected by `withActiveSpan`, or fall back to the propagation context.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/main/integrations/net-breadcrumbs.ts
+++ b/src/main/integrations/net-breadcrumbs.ts
@@ -13,6 +13,7 @@ import {
   setHttpStatus,
   startInactiveSpan,
   stringMatchesSomePattern,
+  withActiveSpan,
 } from '@sentry/core';
 import { logger } from '@sentry/node';
 import type { ClientRequest, ClientRequestConstructorOptions, IncomingMessage } from 'electron';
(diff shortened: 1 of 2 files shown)


lcian:
I will now add tests, I'm not sure of the quality of such tests, so feel free to accept/reject them if it's slop.
That said, I've verified the before and after using a small repro app and this behaves correctly in all cases I could think of.

timfish (approved):

PR #1400, opened 2026-07-08
feat: Span Streaming

timfish:
Tests span streaming mode and ensures and `startupTracingIntegration` works when this mode is enabled.


src/main/normalize.ts
```diff
@@ -185,3 +192,59 @@ export function normalizeProfileChunkEnvelope(
 
   return isProfileChunk ? modifiedEnvelope : envelope;
 }
+
+/**
+ * Normalizes profile_chunk envelope items and returns the modified envelope
```
Lms24: l: This just modifies the span envelope, right?

```suggestion
 * Normalizes span envelope items and returns the modified envelope
```

src/main/normalize.ts
```diff
@@ -185,3 +192,59 @@ export function normalizeProfileChunkEnvelope(
...
+  // Override the SDK info in the envelope header to use Electron SDK
+  const [originalHeader] = envelope;
+  const modifiedHeader = {
+    ...originalHeader,
+    sdk: { name: 'sentry.javascript.electron', version: SDK_VERSION },
+  };
+
+  let modifiedEnvelope = createEnvelope(modifiedHeader);
+  let isSpanContainer = false;
+  let transactionOrigin: string | undefined;
```
Lms24: l: I'd recommend we rename this to `segmentOrigin` so that we don't conflate the "new" spans with the old transaction concept. Would also apply to the consumer of this function. Obviously just logaf-l, so feel free to disregard.

src/main/integrations/startup-tracing.ts
```diff
@@ -155,6 +164,102 @@ function applyRendererSpansAndMeasurements(parentSpan: Span, event: Event | unde
...
 }
 
+function streamedAttr(span: SerializedStreamedSpan, key: string): string | undefined {
+  return span.attributes?.[key]?.value as string | undefined;
+}
+
+// Attributes on the streamed pageload segment that describe the renderer segment itself (its
+// identity and SDK metadata) rather than the measurements and trace metadata (Web Vitals,
+// connection/device info, etc.) that should be merged onto the startup span.
+const NON_INHERITED_SEGMENT_ATTRIBUTES = new Set([
```
Lms24: l: We can use `@sentry/conventions/attributes` to avoid the hard-coded strings and get notified about attribute deprecations. Not directly related to span streaming but IMHO a small win over the hard coded string list here. However, we still need to keep the list itself around so it's not a huge win... feel free to disregard or do as a follow-up. Just a recommendation.

src/main/integrations/normalize-paths.ts
```diff
@@ -22,5 +22,11 @@ export const normalizePathsIntegration = defineIntegration(() => {
     processEvent(event) {
       return normalizePaths(event, app.getAppPath());
     },
+    processSpan(span) {
+      span.name = normalizeUrlToBase(span.name, app.getAppPath());
+    },
+    processSegmentSpan(span, client) {
```
Lms24: I think we can get rid of the `processSegmentSpan` hook here: All spans go through `processSpan` including segment spans.

test/e2e/test-apps/span-streaming/custom-tracing/src/main.js
```diff
@@ -0,0 +1,23 @@
...
+const { app } = require('electron');
+
+const Sentry = require('@sentry/electron/main');
+
+Sentry.init({
+  dsn: '__DSN__',
+  debug: true,
+  tracesSampleRate: 1.0,
+  traceLifecycle: 'stream',
+  integrations: [Sentry.spanStreamingIntegration()],
```
Lms24: l: I think we don't need to set `spanStreamingIntegration()` explicitly anymore, right? All clients except for BrowserClient now add the integration, depending on `traceLifecycle`.

(If you want to keep the test as-is to cover users still explicitly adding `spanStreamingIntegration`, that's fine, too! no objections!)

src/main/integrations/startup-tracing.ts
```diff
@@ -155,6 +164,102 @@ function applyRendererSpansAndMeasurements(parentSpan: Span, event: Event | unde
...
+  'sentry.span.source',
+  'url.full',
+  'http.request.header.user_agent',
+]);
+
+/**
+ * Merges spans streamed from the renderer (when `traceLifecycle: 'stream'` is used) into the
+ * startup span, mirroring {@link applyRendererSpansAndMeasurements} for the streamed span format.
+ */
+function applyStreamedRendererSpans(parentSpan: Span, spans: SerializedStreamedSpan[], endTimestamp: number): number {
```
Lms24: m/h: What would happen if the pageload span tree gets sent in multiple envelopes? With streaming we have no guarantee that we wait until the pageload segment span ends. So the browser SDK could send child spans in one envelope, and only later more spans (including the segment) in another. We flush every 5 seconds by default, so this could happen for long-running pageload traces.

IIUC, here we stitch up the pageload segment span and its children into the startup span trace (`parentSpan`). So previously sent child spans might reference the wrong pageload parent span, since we start a new one here, no?  

I might be missing something though that makes this not become an issue 😅. If it is indeed an issue, I wonder if we could buffer incoming span envelopes until we get the `pageload` segment. Then, we ignore/forward all buffered spans that are not part of the `pageload` trace as-is, and then merge all `pageload`-trace spans together before we hand them to `applyStreamedRendererSpans`. WDYT?

Lms24 (commented):
This looks mostly good to me, thanks for taking this on!

Two comments we should address:
- the timing issue when merging startup and pageload spans
- do we still need users to set `spanStreamingIntegration()`?

All other comments are nits/recommendations, so feel free to disregard them/follow-up on in later PRs.

src/main/ipc.ts
```diff
@@ -136,11 +197,45 @@ function handleEnvelope(
...
+      // Buffered spans from other traces are forwarded unmodified once the wait ends.
+      if (bufferedSpanEnvelopes) {
+        bufferedSpanEnvelopes.push(normalizedSpanEnvelope);
+
+        if (segmentOrigin === 'auto.pageload.browser' && ipcMainHooks.listenerCount('pageload-spans') > 0) {
+          const pageloadTraceId = spanContainerFromEnvelope(normalizedSpanEnvelope)?.items.find(
+            (span) => span.is_segment,
+          )?.trace_id;
+
+          ipcMainHooks.emit('pageload-spans', flushSpanEnvelopeBuffer(pageloadTraceId), contents);
```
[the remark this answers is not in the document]
timfish: Verdict: not a real bug

src/main/ipc.ts
```diff
@@ -96,6 +98,57 @@ function captureEventFromRenderer(
...
+    const container = spanContainerFromEnvelope(envelope);
+
+    if (!container) {
+      continue;
+    }
+
+    if (extractTraceId) {
+      extracted.push(...container.items.filter((span) => span.trace_id === extractTraceId));
+      container.items = container.items.filter((span) => span.trace_id !== extractTraceId);
+    }
```
Lms24, line 135: super-l: Span envelopes always only carry spans of one specific traceId. So, these checks _shouldn't_ be necessary. But they don't hurt either and it's a good safety measure. Happy to leave this up to you.

package.json
```diff
@@ -106,6 +106,7 @@
   },
   "dependencies": {
     "@sentry/browser": "10.67.0",
+    "@sentry/conventions": "^0.16.0",
```
Lms24, line 109: super-l: 0.19.0 is the latest version but if you wanna stay in lockstep with v10 of the JS repo, 0.16.0 is fine.

Lms24 (approved):
Nice, thanks for applying my suggestions! The buffering approach now looks correct to me.

PR #1421, opened 2026-09-01
feat: Emit OTel-aligned http.client attributes from the net integration

timfish:
Stacked on top of #1415 (`feat/vNext`).

Aligns the `ElectronNet` outgoing-request span with the `http.client` spans that `@sentry/node`'s fetch instrumentation emits, so that under span streaming Relay can infer the low-cardinality span **name** and the high-cardinality **description** from the same [sentry-conventions](https://getsentry.github.io/sentry-conventions/) rules.

### What changed

`src/main/integrations/net-breadcrumbs.ts`:
- **Name** is now the sanitized URL — `` `${method} ${sanitizedUrl}` `` via `getSanitizedUrlStringFromUrlObject` (client spans don't branch the name; Relay derives both name and description from attributes).
- **Attributes** are now OTel-aligned: `http.request.method`, `server.address`, `server.port`, `url.full`, `url.path`, `url.query`, `url.scheme` (URLs/queries filtered for PII), replacing the bespoke `url` / `http.method` / `type: 'net.request'`.
- **`onlyIfParent`** now follows `hasSpanStreamingEnabled(client)`, so the span is emitted without an active parent under span streaming — matching the node/browser fetch instrumentation. Correct for both `traceLifecycle: 'stream'` (default) and `'static'`.
- The Electron-specific `sentry.origin` (`auto.http.electron.net`) is kept.

This mirrors `@sentry/node`'s undici/fetch instrumentation, minus the origin and the socket-derived extras (`network.peer.*`, `http.request.method_original`) which aren't needed for name/description inference.

### Scope

SDK side only (the emitting integration). The `electron.*` startup ops and forwarded renderer spans are untouched — the startup ops are already static low-cardinality, and renderer spans are owned by `@sentry/browser`. Verifying/adding the `http.client` rules lives in `getsentry/sentry-conventions`.


src/main/integrations/net-breadcrumbs.ts
```diff
@@ -103,10 +109,49 @@ type RequestOptions = string | ClientRequestConstructorOptions;
...
+      attributes['server.address'] = parsed.hostname;
+      attributes['url.scheme'] = parsed.protocol.replace(/:$/, '');
+
+      if (parsed.port) {
+        attributes['server.port'] = Number(parsed.port);
+      }
+    }
+  }
+
+  return { name: `${method} ${sanitizedUrl}`, attributes };
```
Lms24: m: when streaming is enabled, the name should follow these rules: https://getsentry.github.io/sentry-conventions/names/#http-client

meaning the `sanitizedUrl` should be replaced with only the domain. Relay will use `url.full` to still infer a more detailed description (https://getsentry.github.io/sentry-conventions/descriptions/).

src/main/integrations/net-breadcrumbs.ts
```diff
@@ -103,10 +109,49 @@ type RequestOptions = string | ClientRequestConstructorOptions;
...
+ * and high-cardinality description from these via the conventions rules.
+ */
+function getSpanDetails(method: string, url: string, client: Client): { name: string; attributes: SpanAttributes } {
+  const parsed = parseStringToURLObject(url);
+  const sanitizedUrl = parsed ? getSanitizedUrlStringFromUrlObject(parsed) : url;
+
+  const attributes: SpanAttributes = {
+    [SEMANTIC_ATTRIBUTE_SENTRY_ORIGIN]: 'auto.http.electron.net',
+    'http.request.method': method,
+    'url.full': filterCollectedUrl(url, client),
```
Lms24: logaf-l but wdyt about using `@sentry/conventions` and importing the attribute names from there? It's what we've been migrating to in the JS SDK repo

